PR #988, opened 2026-06-06
fix(test): bound vitest fork memory by native peak + single-runner mutex (T11860)

kryptobaseddev:
## The freeze persisted despite T11839 — here's why, with measurements

The prior cap (worker ceiling **6**, `--max-old-space-size=4096`) still OOM-froze a 24-core/62GB box. Root causes found 2026-06-06:

1. **`--max-old-space-size` caps only the V8 heap, not native memory.** Measured per-fork peak = **~6.2 GB** (4 GB heap + ~2.2 GB native node:sqlite + vec0), not the ~2.7 GB the original estimate assumed. At ceiling 6 → ~39 GB; with a **second concurrent agent** running the suite + other apps → 99% → freeze.
2. **`systemd-run --user --scope -p MemoryMax` is a no-op here.** The user manager has `Delegate=no`, so `memory.max` is never set on the leaf scope — the "mandated" cgroup cap enforced nothing. Verified by reading the scope's empty `memory.max`.

## Fix
- **`vitest.config.ts`**: per-fork RAM budget 6→7 GB (real peak), hard worker **ceiling 6→4** → hard **~26 GB** bound. CI (2-4 cores) unaffected (1-2 forks).
- **`scripts/safe-test.sh`** (new): cgroup-independent runner — `flock(/tmp/cleo-vitest.lock)` so **only one vitest runs machine-wide** (kills the two-agents-at-once freeze) + a `MemAvailable` watchdog that kills the run before the host freezes.
- **`package.json`**: `pnpm test:safe` uses the wrapper.

Validated: wrapper acquires the lock, runs green, tracks peak. Single-process bound is now hard; cross-process collisions are prevented by the mutex.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "pretypecheck": "node scripts/preemit-skill-root-dts.mjs",
     "typecheck": "tsc -b",
     "test": "vitest run",
+    "test:safe": "bash scripts/safe-test.sh -- vitest run",
     "test:watch": "vitest",
     "test:pkg": "vitest run --project",
     "test:changed": "pnpm --filter \"...[HEAD~1]\" run test",
--- a/scripts/safe-test.sh
+++ b/scripts/safe-test.sh
@@ -1,0 +1,69 @@
+#!/usr/bin/env bash
+# safe-test.sh — memory-safe, single-runner vitest wrapper (T11860).
+#
+# WHY: `--max-old-space-size` (vitest.config.ts) caps only the V8 heap, not a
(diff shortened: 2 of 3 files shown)


kryptobaseddev:
Closing as superseded. T11860 (vitest fork OOM cap) shipped via #973 with the maxWorkers=min(cpus-1,RAM/6,6)+4GB-heap-cap approach; T11860 is marked done. This PR's native-peak+mutex alternative is stale (60+ commits behind, mergeable=UNKNOWN). Reopen if the mutex angle is still wanted as a follow-up.

PR #993, opened 2026-06-07
feat(store): E4 — backfill 672 stranded provenance rows bare→prefixed with DHQ-068 coercions (T11883)

kryptobaseddev:
## What

E4 of the complete-cutover (epic **T11883** · saga **T11242**). Recovers the provenance rows that exodus stranded in the bare tables so they survive into the prefixed consolidated substrate.

The exodus bare→prefixed copy strands **672** otherwise-valid provenance rows because a few CHECK-violating legacy values abort their batches (DHQ-068):
- `commits.conventional_type = 'style'` (1)
- `task_commits.link_source = 'commit-message'` (97 — the same bug E3 fixed in reconcile.ts)

This migration `INSERT OR IGNORE`s every stranded bare row into its prefixed table, coercing the two out-of-enum columns (`'style'`→NULL, `'commit-message'`→`'commit-subject'`) so the restored T11363 CHECKs pass. Parents-first ordering + `WHERE EXISTS` FK guards on the release-child tables. Idempotent.

## Dry-run (on a 695 MB VACUUM-INTO clone of the live cleo.db)
- stranded **672 → 0** (commits +101, commit_files +285, task_commits +97, release_commits +175, releases +2, release_changes +8, release_artifacts +2, brain_release_links +2)
- **0 out-of-enum residue** in both coerced columns; `integrity_check` ok; re-apply is a no-op.
- (Pre-existing, unrelated: 2 orphan `tasks_task_relations` FK rows — separate cleanup.)

## Scope note — the bare-table DROP (E5) is deferred
Runtime `foreign_keys = ON` + the deferred satellites (`audit_log`, `token_usage`, lifecycle/agent/experiments/pipeline_manifest/warp_chain_*) are **still bare-bound** and hold recent live-written rows. Safely dropping bare `tasks` requires completing the full satellite rebind + live-wire backfill first — tracked as a follow-up. This PR ships only the safe, additive provenance backfill.

Epic: T11883 · Saga: T11242

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/packages/core/migrations/drizzle-cleo-project/20260607000400_t11883-e4-backfill-stranded-provenance/migration.sql
+++ b/packages/core/migrations/drizzle-cleo-project/20260607000400_t11883-e4-backfill-stranded-provenance/migration.sql
@@ -1,0 +1,134 @@
+-- T11883 (E4) — Backfill stranded provenance rows from the BARE legacy tables
+-- into the PREFIXED consolidated tables, with DHQ-068 enum coercions.
+--
+-- ## Why this exists
+--
+-- The exodus migration copies the bare provenance tables (`commits`,
+-- `task_commits`, `releases`, …) into their prefixed equivalents (`tasks_commits`,
+-- …). The prefixed tables carry the restored T11363 CHECK constraints that the
+-- no-CHECK legacy tables never enforced (DHQ-068). A handful of legacy rows hold
+-- values OUTSIDE those enums:
+--   • `commits.conventional_type = 'style'`        (1 row — not a CC enum member)
+--   • `task_commits.link_source  = 'commit-message'` (97 rows — the same bug E3
+--     fixed in reconcile.ts; the valid member is 'commit-subject')
+-- Because exodus copies in batched transactions, those CHECK-violating rows abort
+-- their whole batch, stranding ~672 otherwise-valid rows in the bare tables. E5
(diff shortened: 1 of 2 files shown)


kryptobaseddev:
Deferring E4 (folding into T11887, the E5 follow-up).

**Why:** E4-as-a-drizzle-cleo-project-migration runs on EVERY consolidated `cleo.db`, but its `INSERT OR IGNORE INTO tasks_X SELECT … FROM <bare>` references the BARE legacy tables, which exist **only on the live DB** that exodus migrated. On a fresh `cleo.db` (new install / test fixtures) there are no bare tables → `DrizzleError: no such table: releases` → migration fails. (Verified: it breaks the reconcile-v2/backfill fixtures once the statement-breakpoints make the runner actually execute the INSERTs.)

**Why this is safe:** the 672 stranded provenance rows live in the bare tables, which aren't dropped until E5 — so nothing is lost. The proper home for this coercing backfill is **exodus's conditional bare→prefixed copy** (which already guards on bare-table existence and is where DHQ-068 enum-coercion belongs), to be done with the full satellite rebind in **T11887**.

The verified-on-clone migration SQL (stranded 672→0, coercions correct, idempotent) is preserved on this branch (`9e0c54fb1`) as the reference for T11887. **Shipping E2+E3 (already on main) via E6 still delivers the DHQ-051 fix to the installed binary.**

PR #1150, opened 2026-08-05
fix(caamp): heal damaged CAAMP markers instead of duplicating blocks (T12051) — includes #1149

kryptobaseddev:
Fixes **T12051**. A CAAMP marker that loses a single character used to be unrecoverable *and* self-amplifying, and the repair `cleo doctor` prescribed for it made things strictly worse.

## The defect

The user's global hub `~/.agents/AGENTS.md` was found holding **three** CAAMP blocks. Two of them had a mangled opening marker — `!-- CAAMP:START -->`, the leading `<` gone:

```
<!-- CAAMP:START -->        ← well-formed
@~/.cleo/templates/CLEO-INJECTION.md
<!-- CAAMP:END -->

!-- CAAMP:START -->         ← damaged
@~/.cleo/templates/CLEO-INJECTION.md
<!-- CAAMP:END -->
                            (and again)
```

By the end of one working session the same file held **thirteen**.

### The ratchet, reproduced

`inject()` matched only complete `START…END` pairs. A damaged opening marker was therefore invisible, so the function concluded the file had *no* CAAMP block and took its **prepend** branch — adding a whole new block beside the damaged one instead of replacing it. Running the built injector against a healthy file reproduces the user's file byte for byte:

```
step 1: fresh inject          → created    { good: 1, damaged: 0, ends: 1 }
step 2: re-inject (idempotent)→ intact     { good: 1, damaged: 0, ends: 1 }
step 3: eat one "<"           →            { good: 0, damaged: 1, ends: 1 }
        inject #1             → added      { good: 1, damaged: 1, ends: 2 }   ← ratchet
        inject #2             → intact     { good: 1, damaged: 1, ends: 2 }
step 4: eat another "<", inject→ added     { good: 1, damaged: 2, ends: 3 }   ← the exact observed file
```

And the loop had no exit: `checkInjection` reported *"markers unbalanced"* with `fix: 'cleo upgrade'` — but `cleo upgrade` calls `ensureInjection` → `inject()`, which is what produced the duplicate. The advertised repair could only ever add another block.

### Why only this file

A filesystem sweep found every project file and every other provider's global file at exactly 1 START / 1 END. Only the global hub was corrupt — and it is the one file **every** `cleo init`, `cleo upgrade` and `cleo doctor` run rewrites, no matter which project invoked them.

The duplication source turned out to be measurable rather than theoretical: `ensureInjection()` resolves the hub through `getAgentsHome()`, and **no test ever overrode it**, so every run of the injection suites wrote the developer's real `~/.agents/AGENTS.md`. T9020 had already diagnosed this ("37 stale blocks observed in production") but only fixed the *content* of the stray blocks so dedup-by-content would collapse them — it never stopped the write. The moment a marker was damaged, dedup-by-content stopped matching and accumulation resumed.

Brute-forcing every marker-touching transformation in the codebase against eleven file states produced **no** malformed marker, and the file is 230 bytes of pure ASCII with no NULs — so the single-byte loss itself came from outside CLEO (an editor, a hook, a heredoc). What this PR fixes is that CLEO could not **detect**, could not **repair**, and actively **amplified** it.

## The fix

**Grammar SSoT** — `packages/contracts/src/caamp-markers.ts`. The marker grammar was re-declared in five places, each recognising a slightly different dialect. It now lives in the leaf package (const data only, so the contracts-purity gate still passes); `core` and `caamp` depend on each other, so neither could own it. The `'created' | 'added' | …` action union, previously inline in eight locations, is now `CaampInjectionAction`.

**Marker engine** — `packages/caamp/src/core/instructions/markers.ts`. `normalizeMarkers()` heals near-miss markers to canonical form *before* any decision is made about the file. `reconcile()` is pure: heal → replace the **first** block in place → drop the rest → preserve everything outside the markers.

Damage tolerance is deliberately bounded. Comment punctuation must be present on **both** sides of the token, and the token must be the whole line. An earlier, looser pattern accepted a bare `CAAMP:START`, which rewrote a documentation code-fence that merely *mentioned* CAAMP into a real marker and swallowed the fenced body — caught by adversarial review, with a regression test.

**Atomic writes + real locking** — `packages/caamp/src/core/fs/atomic.ts`. `writeFileAtomic` is tmp-then-`rename(2)`. `withFileLock` is an `O_EXCL` guard carrying a **fencing token**, so a holder whose guard was reclaimed as stale cannot delete the *next* holder's guard — a mutual-exclusion violation demonstrated at 5.2 s in review. Stale reclaim went 5 s → 30 s and the acquisition budget 1 s → 10 s, both of which fired under exactly the contention the lock exists for. `lock-utils.ts` was generalised onto these instead of keeping its own copy.

**`inject()` fails closed.** A read returning 0 bytes for a file that is non-empty on disk is a torn read from a foreign non-atomic writer; reconciling from it would replace the user's whole file with a lone block. It now throws and leaves the file untouched.

**Repair that can satisfy its own diagnosis** — `cleo caamp repair [--dry-run]`, covering the project files, the global hub, and all 23 detected provider files. It merges blocks whose bodies *differ* rather than only removing identical ones; the previous dedupe-by-identical-content could not fix the "N blocks (expected 1)" state the health check reports, which was another unfixable loop.

**The doctor now looks at the global hub at all**, detects damaged markers and duplicate blocks, and prescribes `cleo caamp repair`.

**Test isolation** — `vitest.setup.ts` pins `AGENTS_HOME` to the per-fork sandbox alongside the existing `CLEO_HOME`/`NEXUS_HOME` pins. Verified: the real hub is byte-identical across a full run of both suites, where it previously grew 3 → 13.

**Two more defects fixed in passing**, both surfaced by the audit:
- `existing.replace(pattern, block)` used `block` as a *replacement string*, so `$&`, `` $` ``, `$'` and `$n` in injected content were substituted — splicing the old block, markers included, inside the new one and forging an unbalanced pair out of ordinary prose. Now the function form.
- `sanitizeCaampFile`'s `/^[A-Z][A-Za-z-]*\.md\s*$/gm → ''` deleted **any** line that was a capitalised filename from the global hub on every bootstrap and every npm postinstall. A user line reading `CONTRIBUTING.md` was silently removed. Those heuristics existed to clean up partially-written blocks, which atomic writes now make impossible, so they are gone.

`pi.ts` carried a second copy of the same ratchet (strict pattern + unconditional append) and now delegates to the shared engine, keeping its append placement via an explicit `insert` option.

## Verification

- `pnpm run build` — green
- `@cleocode/caamp`: **64 files, 1393 passed**, 6 skipped
- `@cleocode/core` injection + doctor suites: **49 passed**
- `cleo check arch` — **7/7 gates pass**; contracts purity gate clean (baseline 53, no net-new)
- `pnpm biome ci .` — 3845 files, exit 0
- **The user's live file was repaired through the shipped engine**: 13 blocks → 1, 1000 → 77 bytes, with every non-block line verified preserved before the write. `cleo caamp repair --dry-run` now reports 23/23 files clean.
- End-to-end through the real CLI on a corrupted sandbox hub: `filesModified 1, markersHealed 1, duplicatesRemoved 1`, and the user's own `# My own notes` / `Keep this line.` survived verbatim.

New regression tests cover every damage class, the exact observed 3-block file, repeated corruption never accumulating, 12-way concurrent injection, torn-read refusal, lock fencing, and repair of distinct-bodied blocks.

### One correction to the original report

The task described this as the ~8k-token protocol being injected 3×. That is **not** what it cost on Claude Code, which deduplicates `@`-imports by resolved path — the `CLEO-INJECTION.md` body appears exactly once regardless of how many references point at it, so the real cost was ~30 tokens of gap text. The defect is a file-integrity and repair-loop problem (unbounded growth, silent user-content deletion, a health check that could never go green), and it does cost tokens on harnesses that do not dedupe.

Task: T12051


kryptobaseddev:
**Supersedes #1149.** This branch was cut from `feat/T12037-project-store-ports`, so it carries #1149's seven E6 ProjectStore commits plus the three CAAMP ones. Merging this lands both; #1149 can then be closed as merged-via-1150.

If you'd rather land them separately, merge #1149 first and this will rebase cleanly onto it.

PR #1159, opened 2026-08-06
fix: the sentient loop spawned workers against sagas and epics (T12079)

kryptobaseddev:
I kept flagging that the loop had never completed *real* work — every sandbox spawn failed because there was no agent runtime. So I gave it a real worker process. That immediately exposed a defect the dry-run path could not.

## The picker had no type filter at all

`saga` and `epic` are **containers**. They group work; they are not work. Spawning a worker against one asks an agent to "do" a grouping, which cannot succeed:

- the attempt burns a retry,
- three burn the item into `stuck`,
- **five stuck items in an hour trip `SELF_PAUSE_STUCK_THRESHOLD`.**

So a handful of unfiltered containers can take the whole autonomous layer offline on its own.

It survived here only by **alphabetical luck** — wave-0 ordering happened to surface `T1009`, a task. In a fresh project the **root saga sorts first** (`T001`), so the very first tick picks a container and fails, every time. Measured on a greenfield sandbox: tick 1 picked the saga; the worker rejected it.

Candidates are now filtered to `task` / `subtask`. An unset type counts as executable, matching `tasks.add`, which infers `task` when type is omitted.

## What the end-to-end run then showed

With a **real** worker — a separate `node worker.mjs <taskId>` that reads the task and edits the repo, not an inline stub:

1. the loop picked the real task **T003**, not the saga;
2. the worker implemented `divide()` with a zero-guard — **the code is in `src/calc.ts`**, real work by a spawned process;
3. the worker tried `cleo update --status done` and was **refused by the ADR-051 evidence gates**;
4. with that removed, the loop *still* returned failure:

```
worker re-verify rejected (T1589/T11498): exit=0 but gates failed
```

**That last line is the finding worth keeping.** The loop does not trust a worker's exit code — it re-verifies gates and refuses to record success for unverified work. `tasksCompleted` stayed `0` while the code change landed. **A worker cannot lie its way to a green tick.** That is a stronger result than a `success` outcome would have been.

## Also found — filed, not fixed (T12079)

`cleo init` **breaks `git add -A` in any fresh project**:

```
error: '.cleo/' does not have a commit checked out
error: unable to index file '.cleo/'
fatal: adding files failed
```

`ensure-dirs.ts` runs `git init` inside `.cleo/`, and git refuses to index an *untracked* directory containing `.git`. This repo is unaffected only because its `.cleo/` already holds **committed** files from before the nested repo existed (verified with `git ls-files` / `git check-ignore`).

I tried a host-`.gitignore` fix and **reverted it** rather than ship something that reads like a fix but isn't:

| attempt | result |
|---|---|
| ignore `.cleo/.git/` | **no** — git still sees the nested repo |
| ignore `.cleo/` wholesale | works, but `removeCleoFromRootGitignore` deliberately strips it so ADRs stay trackable |
| stage `.cleo/.gitignore` | insufficient — needs a *commit*, and init must not commit to your repo |

The real fix relocates the checkpoint repo out of `.cleo/` root and needs a migration for existing installs. Filed rather than rushed.

## Verification

- **35 files / 490 tests** pass in the sentient suite
- 4 new tests covering the container classification and the self-pause arithmetic that makes the filter load-bearing

### DIFF
--- a/packages/core/src/sentient/__tests__/pick-executable-type.test.ts
+++ b/packages/core/src/sentient/__tests__/pick-executable-type.test.ts
@@ -1,0 +1,87 @@
+/**
+ * The sentient picker must skip CONTAINER task types (T12079).
+ *
+ * `saga` and `epic` group work; they are not work. Spawning a worker against
+ * one asks an agent to "do" a grouping, which cannot succeed. The attempt burns
+ * a retry, three burn the task into `stuck`, and five stuck items inside an
+ * hour trip `SELF_PAUSE_STUCK_THRESHOLD` — so a handful of unfiltered
+ * containers can take the whole autonomous layer offline.
+ *
+ * The picker had NO type filter at all. It survived in the CLEO repo only by
+ * alphabetical luck (wave-0 ordering happened to surface `T1009`, a task). In a
+ * fresh project the root saga sorts first — `T001` — so the very first tick
+ * picks a container and fails, every time. Measured on a greenfield sandbox:
+ * tick 1 picked the saga and the worker rejected it.
+ *
+ * @task T12079
+ */
+
+import type { Task } from '@cleocode/contracts';
+import { describe, expect, it } from 'vitest';
+import { runTick } from '../tick.js';
+
+/** Minimal task factory. */
+function task(id: string, type: Task['type']): Task {
+  return {
+    id,
+    title: `${type} ${id}`,
+    status: 'pending',
+    priority: 'medium',
+    type,
+    depends: [],
+  } as unknown as Task;
+}
+
+/**
+ * Drive `runTick` with an injected picker and spawn, capturing which task the
+ * loop chose. `pickTask` bypasses `defaultPickTask`, so these cases assert the
+ * loop's contract; the type filter itself is asserted through the exported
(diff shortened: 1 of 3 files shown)


kryptobaseddev:
## CI blocked by the same GitHub Actions outage as #1158

Every failing check dies at **`Prepare all required actions`**, before any repo code runs:

```
Failed to resolve action download info. Error: Service Unavailable
##[error]Service Unavailable
```

One of them is instructive: **Installer Node Floor SSoT** is marked failed even though its own step *succeeded* in the log —

```
lint-installer-node-floor: OK — both installers embed Node floor 24.16.0 matching engines.node ">=24.16.0"
```

— which confirms the job is being killed by the environment, not by anything in this branch. Verified locally too: that lint exits 0.

### Local verification for this PR

- **35 files / 490 tests** pass in the sentient suite
- 4 new tests (container classification + the self-pause arithmetic that makes the filter load-bearing)
- End-to-end run with a real spawned worker: loop picked task **T003** (not the saga), the worker implemented `divide()`, and the loop then correctly **refused** to record success without evidence gates

**Re-run the failed jobs once the outage clears.** Not merging on red.

PR #1178, opened 2026-08-10
fix: superseded-store diagnosis + hard memory ceiling on heavy tool spawns (T12095, T12096)

kryptobaseddev:
## What happened

An agent working in `/mnt/projects/PepsVida` spent a session going in circles:

> "The current tasks.db may be a different format (maybe the llmtxt.db)… the current tasks.db is 417KB which is much smaller than the backups (58MB). Maybe the current tasks.db was rotated/rebuilt…"

**Nothing was wrong with that project.** Verified: 1,123 tasks, 106 sessions, an active session, 47 pending, `cleo current` → T1094, all readable through the CLI. The on-disk layout manufactured a corruption signal and the agent trusted it over the tool.

## The trap (three leftovers, each individually defensible)

Post-E6 (ADR-068) the store is `.cleo/cleo.db` with task rows in **prefixed** tables (`tasks_tasks`, …).

| Artefact | Reality |
|---|---|
| `.cleo/tasks.db`, 408 KB, last written June 11 | The **pre-migration** store. The migration doesn't delete it, so it survives under the name every doc — including `AGENTS.md` § Runtime Data Safety and `cleo restore backup --file tasks.db` — still uses for "live". |
| `.cleo/backups/sqlite/tasks-<ts>.db`, 58 MB | Snapshots **of `cleo.db`**. `openTasksDbForSnapshot` routes through the dual-scope chokepoint; the `tasks-` prefix is a legacy label kept for the rotation regex. This is the "417KB vs 58MB" read as truncation. |
| A bare `tasks` table inside `cleo.db`, **0 rows** | An empty relic beside the populated `tasks_tasks` (1,123 rows). Any direct SQL probe lands on the decoy. |

The trap is set in **3 projects on this machine**, including `cleocode` itself.

## The fix

`cleo doctor superseded-store` — read-only, deletes nothing — names each superseded file and **proves** which store is authoritative by counting rows in both. Run against the real project:

```
liveStore : /mnt/projects/PepsVida/.cleo/cleo.db
entries   : 1
  • tasks.db (408 KB, modified 2026-06-11)
    rows here: 0   rows live: 1123   safeToArchive: true
```

Two deliberate refusals to guess:

- **Silent when `cleo.db` is absent.** A pre-migration project's `tasks.db` *is* its live store; flagging it would advise deleting the only database.
- **Withholds the archive recommendation** when the legacy file still holds rows, is newer than `cleo.db`, or isn't valid SQLite.

`CLEO-INJECTION.md` — injected verbatim into every spawned agent — now states where the data lives, tabulates all three decoys, says plainly that 408 KB-beside-58 MB is the **normal** layout of a migrated project, and points at the one command. `AGENTS.md` § Runtime Data Safety is corrected from `tasks.db` to `cleo.db` with the same table.

## Verification

- Run against the actual project that triggered this (output above)
- `tsc -b` clean · `biome check` clean · `cleo check arch` **9/9**
- Gate 14 confirms the new command exists (`all 62 referenced commands exist`)
- 6 unit tests cover both directions — correctly flags a migrated leftover, and stays silent on an unmigrated project, garbage file, and newer file
- Suite runs in CI, not locally

### DIFF
--- a/.changeset/t12095-superseded-store-diagnosis.md
+++ b/.changeset/t12095-superseded-store-diagnosis.md
@@ -1,0 +1,16 @@
+---
+id: t12095-superseded-store-diagnosis
+tasks: [T12095]
+kind: fix
+summary: a healthy migrated project read as CORRUPT — three leftovers from the cleo.db migration sent an agent in circles over a 1123-task database that was completely intact
+---
+
+Post-E6 (ADR-068) the project store is `.cleo/cleo.db` with task rows in PREFIXED tables (`tasks_tasks`, …). Three artefacts conspire to say otherwise:
+
+1. `.cleo/tasks.db` — the pre-migration store — is left on disk under the name every doc still uses for "live".
+2. Snapshots are written `tasks-<ts>.db` while actually snapshotting `cleo.db`, so a 408 KB legacy file sits beside 58 MB files bearing its own name.
+3. `cleo.db` keeps a bare, empty `tasks` table next to the populated `tasks_tasks`, so a direct SQL probe finds the decoy.
+
+Measured in a real project: an agent looped on "the current tasks.db is 417KB which is much smaller than the backups (58MB) — maybe it was rotated/rebuilt", then began theorising the store might be `llmtxt.db`. The data was fine — 1,123 tasks, 106 sessions, an active session, 47 pending.
+
+`cleo doctor superseded-store` answers it in one read-only call, naming each superseded file and proving which store holds the data by counting rows in both. It stays silent when `cleo.db` is absent, so it can never tell an operator to delete their only database, and it withholds the archive recommendation when the legacy file still holds rows. The injection template every agent reads now states where the data lives and warns against reasoning about `.db` file sizes.
--- a/.changeset/t12096-heavy-tool-spawn-ceiling.md
+++ b/.changeset/t12096-heavy-tool-spawn-ceiling.md
@@ -1,0 +1,18 @@
+---
+id: t12096-heavy-tool-spawn-ceiling
+tasks: [T12096]
+kind: fix
+summary: one tool:test atom could fan out to 15 unbounded vitest pools in a consuming project — CLEO now sets a hard memory ceiling at the spawn point
+---
+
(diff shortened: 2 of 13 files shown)


kryptobaseddev:
## Added T12096 — the hole T12091 could not see

While investigating T12095 I caught a live process tree from another agent's session that explains the freezes T12091 did **not** fix:

```
cleo verify T1046 --gate testsPassed --evidence tool:test
  └─ npm test
      └─ pnpm -r --if-present run test        ← fans out
          └─ concurrent vitest in /mnt/projects/PepsVida/{apps,lib}/*
```

T12091 bounds how many `tool:test` **invocations** run at once — and it counts that entire tree as **one slot**. PepsVida has **15 workspace packages with a `test` script** and **none of its 3 vitest configs cap `maxWorkers` or heap**: `vitest.memory-safe.ts` and gate 8 live in cleocode and protect only cleocode. So one evidence atom could become 15 concurrent unbounded fork pools while the semaphore reported a single slot in use.

A consuming project can't be relied on to bound its own runner, and CLEO is the process that starts it — so the ceiling now lives at `tool-cache.ts`'s single `spawn`, inherited by the whole tree.

Three levers, each **verified against the installed tool** rather than assumed:

| Variable | Effect | How it was verified |
|---|---|---|
| `NODE_OPTIONS=--max-old-space-size=4096` | Heap ceiling on every node process in the tree | appended, never clobbered — dropping a project's `--experimental-*` flags would break the command |
| `VITEST_MAX_WORKERS=<RAM-derived>` | `if (process.env.VITEST_MAX_WORKERS) resolved.maxWorkers = parseInt(…)` — **overrides** the resolved config, so it binds projects that set their own | read out of vitest 4.1.4 `dist/chunks/coverage.*.js` |
| `npm_config_workspace_concurrency=1` | Bounds the `pnpm -r` fan-out | `npm_config_workspace_concurrency=1 pnpm config get workspace-concurrency` → `1` |

`test`/`build` only. Every value the project set deliberately is preserved.

**Worst case on a 62 GiB box: unbounded × unbounded × unbounded → 1 × 6 × 4 GiB = 24 GiB.**

Also in this push: the earlier CI failure on `superseded-store.test.ts` was a real bug in my code, not a bad expectation. `(rowsInSuperseded ?? 0) === 0` treated an *unreadable* legacy DB as empty and recommended archiving it — the one case where being wrong loses data. Now an explicit `=== 0`.

PR #1251, opened 2026-09-12
fix(T12112): evidence runs in the caller's tree; a tool no longer invalidates its own cache

### LINKED ISSUE #1220: [Bug]:Evidence tools run in the shared repo root, not the worktree the verify was invoked from
### Bug Report

Running `cleo verify <id> --gate qaPassed --evidence "tool:typecheck"` from a git worktree at /home/keatonhoskins/pv-reship executed the tool against /mnt/projects/PepsVida — the main checkout — and failed:

    Tool "typecheck" exited with code 2. Tail: …k /mnt/projects/PepsVida/scripts
    src/_t1525-care-panel-economics.mts(13,44): error TS6059: File
    '/mnt/projects/PepsVida/apps/clinical-api/src/siphox/carePanelConfig.ts' is not
    under 'rootDir' '/mnt/projects/PepsVida/scripts/src'

That file is another agent's UNTRACKED work in the shared checkout. It is not on main, not in my worktree, and not part of what I was verifying. My worktree typechecks clean (exit 0).

Impact: on a machine where several agents share one repo via worktrees, gate evidence is unobtainable for anyone whose peers have a dirty tree — and worse, a PASS could be produced from a tree that is not the one being verified. Evidence that does not describe the code under test is not evidence.

Expected: the tool should run with cwd = the worktree the CLI was invoked from (`git rev-parse --show-toplevel`), or the resolved root should be printed in the envelope so the operator can see which tree was measured.

Repro: create a worktree, dirty the main checkout so a tool fails there, run `cleo verify` from the worktree.

CLEO v2026.8.9 / ADR-061 tool resolution.

**Severity**: Major
**Area**: cli

---

## Environment
| Component | Version |
|-----------|---------|
| CLEO | 2026.8.9 |
| Node.js | v24.16.0 |
| Shell | /bin/bash |

### LINKED ISSUE #1221: [Bug]:Evidence result cache never hits for tool:test, so a batch of verifies re-runs a 10-minute suite per task and stalls
### Bug Report

Gating 19 tasks from one merge required `tool:test` evidence on each. ADR-061 says results are cached on (canonical, cmd, args, HEAD, dirty-tree fingerprint), so one run should cover the batch.

Observed: every `cleo verify … --evidence "tool:test"` re-ran the full suite (~10 min). A sequential loop over 7 tasks produced no output in 40+ minutes and completed none of them. Only the very first task (T1655) ever reached `done`.

Likely cause: the dirty-tree fingerprint component. On a shared box several agents are editing the same checkout continuously, so the fingerprint changes between calls and the key never repeats. Combined with the previous issue (tools run in the SHARED root rather than the caller's worktree), the fingerprint being measured is not even the tree being verified — so the cache is invalidated by other people's edits.

Suggestions:
- exclude UNTRACKED files from the fingerprint, or fingerprint only tracked+staged content;
- fingerprint the worktree the verify was invoked from;
- or offer an explicit `--reuse-evidence <taskId>` / batch verify so one measured run can be attached to N tasks from one merge, which is the common case after a squash merge.

Related: the default per-tool deadline of 300 s is far below a monorepo suite (~10 min here), so the first attempt failed with E_EVIDENCE_TOOL_TIMEOUT before I found CLEO_TOOL_TIMEOUT_TEST. A default derived from the project (or a clearer up-front hint naming the env var) would save the discovery.

CLEO v2026.8.9.

**Severity**: Major
**Area**: cli

---

## Environment
| Component | Version |
|-----------|---------|
| CLEO | 2026.8.9 |
| Node.js | v24.16.0 |
| Shell | /bin/bash |

### LINKED ISSUE #1226: [Bug]:tool: and test-run: evidence resolve against the shared project dir, not the calling worktree
### Bug Report

Running `cleo verify <id> --gate testsPassed --evidence 'tool:test'` from a git worktree executes the project's test command in the SHARED project directory, not in the worktree the caller is working in. Same for `test-run:<path>`: a relative path resolves against the shared tree, so a report written in the worktree reports 'file does not exist'.

Two consequences, both of which make the evidence WRONG rather than merely inconvenient:

1. The captured result attests to whatever branch the shared tree happens to have checked out. In this session that was a peer's in-flight branch carrying a red contract pin, so my change was refused on somebody else's failure. Had that branch been green, I would have recorded a PASS that never ran my code at all. The false-pass direction is the dangerous one and it is silent.

2. Combined with pr: atoms being unusable in a repo without CI, qaPassed (which accepts only tool: or pr:) becomes unreachable for any worktree session with no honest path left, which pushes people toward CLEO_OWNER_OVERRIDE for routine work.

Expected: tool and test-run evidence resolve against the caller's worktree root (git rev-parse --show-toplevel), or at minimum the envelope reports WHICH directory the tool ran in, so the discrepancy is visible rather than inferred from a confusing failure.

Workaround: generate vitest --reporter=json reports in the worktree, copy them into the shared tree's .cleo/agent-outputs/, cite as test-run atoms.

---

## Environment
| Component | Version |
|-----------|---------|
| CLEO | 2026.8.9 |
| Node.js | v24.16.0 |
| Shell | /bin/bash |

### LINKED ISSUE #1230: [Bug]:cleo verify accepts evidence captured from a different tree than the attested commit (tool: runs in the shared checkout; files: can pin a hash no revision ever had)
## Summary
`cleo verify` accepts and records evidence atoms captured from a **different working tree than the commit being attested**, producing green evidence attributed to a commit the run never saw. Two concrete manifestations, both measured on 2026-09-11.

## Manifestation A — `tool:` atoms run in the wrong checkout
This project pins tool commands in `.cleo/project-context.json` as `pnpm --dir axiom-app <script>` because the CLEO root is the PARENT of the git repo. So every `tool:test|lint|typecheck` atom runs in the shared `axiom-app` checkout — while `GIT_DIR`/`GIT_WORK_TREE` (needed to make `commit:` atoms resolve at all) point at the agent's own worktree and supply the HEAD that keys the evidence cache.

At the time of measurement the shared checkout was on `fix/layer4-...` at `4190f4a4` with another session's untracked files, while the attesting worktree was main at `38864d00`. The gate would have recorded "tests passed at 38864d00" from a run that never saw `38864d00`. **It fails silently in the dangerous direction.**

## Manifestation B — a `files:` atom can pin a hash no revision ever had
On task T174 (`verification.passed: true`, all three gates set), **two of three `files:` sha256 pins match NO revision that has ever existed in the repository**:
- `src/lib/intake/vial-queue.ts` pinned at `18d3e8db…`; that file has been byte-identical (`38f9e924…`) since its own merge commit.
- `VialQueue.tsx` pinned at `b34bc4ce…`, matching nothing on any ref.
- The third atom matched main exactly — which proves the hash is plain content-sha256 and the method sound, so this is **not** a hashing-convention artefact.

A green `verification.passed` is therefore not evidence that the pins resolve. CLEO re-validates hard atoms on `complete`, but a hash captured from a tree that later moved is still accepted at capture time.

## Repro
1. In a repo where the CLEO root differs from the git root, pin `tool:` commands to a fixed subdirectory path (as `project-context.json` requires here).
2. From worktree A (commit X), run `GIT_DIR=<A>/.git GIT_WORK_TREE=<A> cleo verify T### --gate testsPassed --evidence "tool:test"` while the pinned directory is checked out at commit Y.
3. The atom is accepted and attributed to X. Nothing reports that the run happened at Y.

## The ask (one cheap check kills the whole class)
**Record `git rev-parse HEAD` at tool-run START and again at END, and refuse the atom when they differ or when either differs from the commit being attested.** This also catches the adjacent failure we both hit by hand: a full-suite run across three different HEADs (720/806 files in) producing output that described no single tree, and a near-duplicate commit from two racing rebases. In both cases the run produced output the whole time and the output meant nothing.

Secondary ask: for `files:` atoms, resolve the recorded hash against the repository at capture time and refuse a pin that matches no blob on any ref.

## Note
`pr:<n>` is not a usable fallback here — it requires CI workflows this repo removed on 2026-09-04 and can never pass.

**Severity**: Blocker
**Area**: cli

---

## Environment
| Component | Version |
|-----------|---------|
| CLEO | 2026.8.9 |
| Node.js | v24.16.0 |
| Shell | /bin/bash |

kryptobaseddev:
Closes #1220
Closes #1221
Closes #1226
Closes #1230

Part of SG-ISSUE-ZERO (T12110) → E-EVIDENCE-PIPELINE (T12111) → T12112.

## What was wrong

Two defects that between them made gate evidence **unobtainable** for anyone sharing a box, and **untrustworthy** when it was obtainable.

### Wrong tree — #1220, #1226, #1230

`getProjectRoot()` deliberately resolves a git worktree to the **main repo**, so every worktree shares one `.cleo/` store. That is correct for the store and wrong as a tool's working directory — and the same value was used for both. A verify launched from a worktree therefore measured the shared checkout, where a peer's in-flight branch and untracked files decided the outcome.

A red peer produced a false FAIL. A **green peer produced a silent false PASS**, attesting a run that never touched the code under test. That direction is the reason this is P1 rather than an inconvenience.

Execution root and store root are now distinct concepts:

- the tool **spawns in**, and is **fingerprinted against**, the caller's tree;
- the cache **entry** stays under the store root — sound because the key is content-addressed, so two trees collide only when they hold the same code;
- the measured tree is **reported on the result** (the reporter's explicit ask), and deliberately on the *result* rather than the shared cache entry, so a hit served to a second worktree cannot claim to have run in the first one's directory;
- `test-run:<path>` resolves relative paths against the caller's tree before falling back to the store root.

Resolution is conservative: the caller's git toplevel is used only when its canonical main repo equals `projectRoot`. Anything else falls back, so **single-checkout behaviour is bit-for-bit unchanged**.

### Cache could never hit — #1221

The reported cause was fingerprint churn from peers editing a shared checkout. **Measured and refuted:** the shared checkout's fingerprint was byte-identical across 6 samples over 60s with 12 live sessions, and `git status --porcelain` there was one line. Two deterministic causes explain the whole report on an idle single-agent box.

**1. The tool invalidated its own cache by running.** The fingerprint hashed `git status --porcelain`, which lists untracked files, so any artifact the tool emitted changed the key the *next* call computes. In this repo `coverage/`, `.vitest/` and `*.log` are **not gitignored** (only `dist` and `node_modules` are). One suite run is enough. A one-line marker file is enough — which is how this was found: a test helper writing its own spawn marker inside the repo under test made a *control* case fail. The fingerprint now covers tracked content only.

**2. The deadline makes caching structurally impossible.** `DEFAULT_SPAWN_TIMEOUT_MS` is 300s; a monorepo suite takes ~10 min. Every run is killed at 5 min, and the timeout path deliberately caches nothing (T12025 — correct: an unfinished run is not a result). So the cache cannot hit, not because the key moved but because **no entry is ever produced**, on a path that always fires. The arithmetic matches the report exactly: 7 tasks × 300s ≈ the reporter's "40+ minutes, completed none of them."

> **This reframes the host impact, and it is the strongest argument for why this cluster was a machine-killer rather than a slowness complaint.** The load was not N redundant cached suites being re-run. It was N suites **guaranteed to be discarded** — each running at full parallelism for the full deadline before being killed. Maximum cost, zero retained value, every time, forever.

The deadline raise ships separately, gated behind the kernel memory bound (a longer deadline must not land before the guardrail). What lands here is the error text: it no longer advises an identical retry — which loops forever at 5 CPU-minutes a turn — and instead names `CLEO_TOOL_TIMEOUT_<CANONICAL>`, the current deadline, and the tree that was measured.

## The tradeoff, stated plainly

Excluding untracked files means an **uncommitted NEW source or test file no longer invalidates the cache either** — so a verify can return a cached pass that did not exercise it. Commit it (HEAD moves, the key moves), or force a measured run with `CLEO_EVIDENCE_FRESH=1`.

This is pinned by a deliberately-named test (`TRADEOFF: a new UNTRACKED source file does NOT invalidate…`) so the next reader meets it as documentation rather than discovering it as a stale pass. The alternative — a maintained per-project exclude list of build-output paths — trades a loud, documented staleness for a silent, per-project one that rots as tooling changes.

## Verification

No unscoped suite was run at any point (this PR is about a tool that freezes hosts).

| Scope | Result |
|---|---|
| `tool-cache` + `evidence` + `tool-semaphore` + new | **128 passed** |
| remaining evidence-related core suites (8 files) | **227 passed** |
| `@cleocode/worktree` spawn-verify e2e | **3 passed** |
| `@cleocode/cleo` verify-explain | **13 passed** |
| `tsc --noEmit` across `@cleocode/core` | **clean** |

8 new cases cover both causes, the tradeoff and its escape hatch, worktree execution, the `executionRoot` default, and the #1230 false-PASS direction.

Two existing tests encoded the old untracked-file semantics. They were **rewritten to assert the new intent with rationale, not flipped** — and split so that tracked-file invalidation keeps explicit coverage, guarding against over-correcting the fix into a cache that is blind to real edits.

## Note for reviewers

`runToolCached`'s `spawnCmd` call site moved in this PR (`projectRoot` → `executionRoot`). Any patch applied against the old text will silently revert #1220 while leaving every test green — because the tests would still pass, just measuring the wrong tree again. That is the failure mode this PR exists to remove, so please re-read that call site rather than trusting a clean merge.


kryptobaseddev:
## Note on `tool-cache.test.ts > "returns lockBusy:true within 4 s when the cache lock is held externally"`

If you see this assertion fail, **it is not introduced by this PR** — it is environmental, and it depends on how your worktree's `node_modules` is provisioned. Recording the control set so nobody re-derives it:

| tree | `node_modules` | result |
|---|---|---|
| this branch `86ade53e8` | symlinked | FAIL (5.5–8.9s) |
| `main` 77fb47e04, fresh worktree | symlinked | **FAIL (7.8s)** |
| `main` 77fb47e04, shared checkout | real install | PASS |
| **this branch, real `pnpm install`** | real install | **PASS (135/135)** |

Rows 2 and 4 are the ones that settle it. Main's own unmodified code fails the same assertion under a symlinked worktree, and this branch passes it under a real install. The variable is module resolution, not the diff.

Credit where due: the first three rows pointed squarely at this PR and were wrong. @cleo-dev ran the fourth control rather than filing the regression.

All verification in the PR description has been **re-run against a real `pnpm install`**, not the symlinked tree: 135 passing across `tool-cache`, `tool-semaphore`, `evidence` and the two new suites.

The 4-second budget in that test is worth revisiting separately — an assertion that flips on `node_modules` layout is measuring the machine more than the code — but that is out of scope here and filed on its own.

kryptobaseddev:
## CI fix: `Project Root Anti-Pattern Lint` — and why the suggested fix was wrong

The lint flagged `cwd: string = process.cwd()` in `resolveEvidenceExecutionRoot` and suggested `resolveOrCwd(...)`. **Taking that suggestion would have silently reverted this entire PR.**

`resolveOrCwd(maybeRoot)` falls back to `getProjectRoot()`. And `getProjectRoot()` deliberately collapses a worktree to its main repo — *that collapse is the bug this PR exists to undo*. Routing through it makes `resolveEvidenceExecutionRoot` return `projectRoot` unconditionally: evidence goes back to being measured in the shared checkout, the silent false-PASS returns, and **nothing fails**. Every test stays green, because the tests would simply be measuring the wrong tree again — which is the exact failure mode this PR removes.

A trap with no observable signature. Recording it here so it isn't re-introduced by a future reader treating the lint as authoritative.

**The lint's rule is correct in general**: a bare `process.cwd()` used as a stand-in for the project root is a real defect, and that is what the rule was written to catch. This is the inverse case — the invocation directory is the **subject** of the function, not a proxy for anything else. It is also used only to *locate a candidate*; the same-project check immediately below decides whether that candidate is honoured, and falls back to `projectRoot` for anything that is not a worktree of this project. Single-checkout behaviour is unchanged.

Annotated with the sanctioned `// CWD-OK: <reason>` marker rather than allowlisting the file, so the reasoning is at the line as well as here. `lint-project-root-anti-pattern --strict`: **zero violations**. 103 tests still green across evidence and tool-cache.

**Mechanical note for anyone hitting this lint:** the marker must be a **trailing comment on the offending line**. A rationale block placed immediately *above* the line still reads as unannotated and the lint keeps failing — which is easy to misread as the lint being flaky.

Issue #1254, opened 2026-09-12
[Bug]:tool-cache lockBusy test has a 4s wall-clock budget that flips on node_modules layout, not on code


kryptobaseddev:
### Summary

`packages/core/src/tasks/__tests__/tool-cache.test.ts` → `"returns lockBusy:true within 4 s when the cache lock is held externally"` asserts a wall-clock budget. That budget is sensitive to module-resolution cost, so the test's result is decided by how the working tree was provisioned rather than by the code under test.

### Evidence

Measured 2026-09-12 across four trees, holding the assertion constant:

| tree | `node_modules` | result |
|---|---|---|
| `main` 77fb47e04, shared checkout | real install | PASS |
| `main` 77fb47e04, fresh worktree | symlinked | **FAIL (7.8s)** |
| feature branch, worktree | symlinked | FAIL (5.5–8.9s) |
| feature branch, worktree | real install | PASS |

Row 2 is the important one: **unmodified `main` fails its own test** when `node_modules` is a symlink to a shared store rather than a real install.

### Why it matters

This cost two sessions real time during SG-ISSUE-ZERO. The first three data points available pointed at an innocent feature branch, and the regression was nearly filed against it; only running `main` under the same provisioning showed the branch was clean. A test that can accuse an unrelated diff is worse than no test on that dimension.

It also means **neither a pass nor a failure of this assertion is evidence** in a worktree-based workflow — which is the workflow CLEO itself provisions for agents via `cleo orchestrate spawn`.

### Suggested shape

The thing worth asserting is *that the call returns `lockBusy: true` promptly rather than blocking for the full lock timeout* — an ordering property, not a duration. Options, roughly in preference order:

1. Assert the outcome (`lockBusy === true`) and that it returned before the lock's own stale/timeout window, deriving the bound from that window instead of a literal `4000`.
2. Inject a clock / fake timers so the assertion is deterministic.
3. If a wall-clock budget must stay, set it from a measured baseline taken in the same process, so it scales with the machine and the module-resolution cost.

### Related

Not a blocker for any current cluster — filing rather than fixing mid-saga, per triage.

CLEO v2026.8.9.

Issue #1241, opened 2026-09-11
[Bug]:memory observe / add: process killed AFTER the write commits, so a retry would duplicate


kryptobaseddev:
### Bug Report

**SECOND REPRO, and it sharpens #1237.** Same exit-143-with-no-output, but this time I checked afterwards and found **the write had succeeded**.

```
cleo add --parent T124 --type task --kind bug --severity P2 --title "..." --acceptance "a|b|c|d|e"
# killed at 200s, exit 143, NO stdout, NO stderr

cleo find "unbounded selects that 507"
# T284  pending  medium  18 scripts are dead now and 15 more on the same path...
```

**The task was created. Only the response was lost.**

**WHY THIS IS WORSE THAN A TIMEOUT.** In #1237 I reported a BRAIN observation as filed when it had NOT been written. Here the opposite: the write landed and the caller cannot know. **The two failures are indistinguishable from outside**, and they call for opposite recoveries:

- believe it failed → retry → **duplicate task/observation**
- believe it succeeded → **silently lost write** (the #1237 case)

An agent following the obvious rule ("non-zero exit means retry") creates duplicates in exactly this case. I only avoided it by running `cleo find` first, which is not a step anyone would think to take after a command that printed nothing at all.

**ASK**
1. Emit an envelope on signal termination — even `{"success":false,"error":{"code":"E_KILLED","committed":<bool>}}` — so a caller can distinguish "wrote then died" from "died before writing". `committed` is the field that matters; without it the exit code is not actionable.
2. Failing that, make the mutation **idempotent on a caller-supplied key**, so a retry is safe regardless.

**ENV** axiom-analytics · cleo v2026.8.9 · Linux · heavy multi-agent load (80 concurrent vitest processes measured on this box) · commands wrapped in `timeout 200`

### Are you using an AI agent?

Yes - AI agent filed this issue

---

## Environment
| Component | Version |
|-----------|---------|
| CLEO | 2026.8.9 |
| Node.js | v24.16.0 |
| OS | linux 7.1.5-201.fc44.x86_64 x64 (x64) |
| Shell | /bin/bash |
| gh CLI | gh version 2.97.0 (2026-07-31) |
| Install | /home/keatonhoskins/.npm-global/bin/cleo |

kryptobaseddev:
## Partially addressed by #1258 — the cause, not the contract

Recording precisely what changed, because it would be easy to read #1258 as closing this and it does not.

[...]

Issue #1222, opened 2026-09-05
[Bug]:Tool semaphore is not released when a verify's parent shell dies — E_EVIDENCE_TOOL_BUSY blocks every later verify


kryptobaseddev:
### Bug Report

After several backgrounded `cleo verify … --evidence "tool:test"` invocations were interrupted (their parent shells were terminated), the node processes survived as orphans and kept holding slots under ~/.local/share/cleo/locks/tool-test/.

Every subsequent verify then failed immediately with:

    E_EVIDENCE_TOOL_BUSY

There is no way to see who holds the slot and no documented way to reclaim it. I only recovered by finding the orphans with `ps aux | grep 'cleo verify'` and `kill -9`-ing my own PIDs. On a shared machine that is dangerous advice — the next person to hit this may kill a slot holder that is legitimately running.

Requests:
- release the slot when the holder's process group dies (flock, or record the PID and treat a dead PID as a stale slot);
- make E_EVIDENCE_TOOL_BUSY name the holding PID/start time and the lock path, so the operator can tell 'genuinely running' from 'orphan';
- add `cleo doctor tool-locks` (or similar) to list and reap stale slots.

CLEO v2026.8.9 / ADR-061.

**Severity**: Major
**Area**: cli

### Are you using an AI agent?

Yes - AI agent filed this issue

---

## Environment
| Component | Version |
|-----------|---------|
| CLEO | 2026.8.9 |
| Node.js | v24.16.0 |
| OS | linux 7.1.5-201.fc44.x86_64 x64 (x64) |
| Shell | /bin/bash |
| gh CLI | gh version 2.97.0 (2026-07-31) |
| Install | /home/keatonhoskins/.npm-global/bin/cleo |

Issue #1207, opened 2026-08-20
[Feature]:cleo show is ~6s per call and there is no batch lookup — a 25-task status sweep takes 2.5 minutes


kryptobaseddev:
## What happened

Reconciling a session's work needs the status of a known set of task IDs. There is no way to ask for more than one at a time, so the only option is a shell loop:

```bash
for t in T1425 T1426 ... T1465; do
  cleo show $t --field /data/task/status
  cleo show $t --field /data/task/title
done
```

50 invocations at ~6s each = **over 4 minutes of wall clock**, which exceeded the agent harness's 2-minute command timeout twice and had to be re-run as a background job. Each invocation pays full process startup + DB open + LAFS envelope construction to return one string.

## Why `cleo find` / `cleo list` do not cover it

- `cleo find "query"` is a search over text — it cannot take a list of IDs, and it does not reliably return every ID asked for.
- `cleo list --parent <id>` returns children of one parent. The IDs in a session's work span many parents (they are "what I touched", not "one epic").
- `cleo list` without `--parent` is explicitly discouraged in the protocol for token reasons.

So a caller who knows exactly which N tasks it wants has no cheap path, which is the common case at session end.

## What would fix it

Accept multiple IDs on `show`, and return one envelope:

```bash
cleo show T1425,T1426,T1427 --output table
cleo show T1425 T1426 T1427 --summary
```

Either shape works. `--summary` already emits one line per record, so a multi-ID `show --summary` would be an ideal reconciliation surface. A `--status-only` projection would be even cheaper.

Failing that: a `cleo status <id...>` command that skips the render layer entirely.

## Secondary observation

`cleo show <id> --field /data/status` fails with `E_FIELD_NOT_FOUND` because the record is nested under `task`. The error message is genuinely good — it lists the valid pointers and explains the nesting — but the mutate commands documented in the CLI output contract use un-nested paths (`/data/created/0`, `/data/count`), so `/data/status` is the natural first guess. Worth a line in the contract docs noting that `show` nests and mutations do not.

## Environment

CLEO v2026.8.7, project PepsVida (1400+ tasks, 82 MB cleo.db), Fedora, local SQLite backend.

### Are you using an AI agent?

Yes - AI agent filed this issue

---

## Environment
| Component | Version |
|-----------|---------|
| CLEO | 2026.8.7 |
| Node.js | v24.16.0 |
| OS | linux 7.1.5-201.fc44.x86_64 x64 (x64) |
[...]

PR #1258, opened 2026-09-12
fix(T12115-T12117): a cleo command must finish, exit, and never lose a write

### LINKED ISSUE #1228: [Bug]:cleo verify/update/complete do not exit after printing their success envelope
## Observable

`cleo verify`, `cleo update` and `cleo complete` print their success envelope, commit their write, and then **never exit**. The work succeeds; the process stays alive indefinitely.

Cause deliberately not asserted — see "What I did not determine" below.

## Measured

`ps -eo etime,args` on one machine, 2026-09-10, three agent sessions on one repo. Every one of these had already printed `{"success":true,...}` and its effect was verified by a separate `cleo show`:

```
01:41  cleo verify   T1712 --gate testsPassed --evidence test-run:/tmp/t1712-conferencing.json
01:30  cleo update   T1035 --blocked-by "..." --output silent
01:29  cleo verify   T1812 --gate implemented --evidence commit:8d727c0c...
01:15  cleo complete T1712
   46  cleo verify   T1815 --gate implemented --evidence commit:2cbb0036...
```

`cleo show T1812` reads `implemented/testsPassed/qaPassed` all true and `status: done`. So the 1h29m process finished its job in seconds.

**Three different subcommands**, so this is below the subcommand layer. A `commit:` atom is a git reachability check plus a file sha256 — it is not slow, it simply does not terminate.

## Impact, which is larger than the wall clock

⛔ **`$(cleo ...)` blocks until EOF, not until exit status.** Command substitution waits for the pipe to close, which needs the process to die. So a shell loop over N tasks does not take N x 12s — it stalls on the first lingerer. Measured: a completion loop over 21 tasks produced 2 results in 10 minutes, then ran normally once the workaround was applied.

⛔ **It reports SUCCESSFUL work as failed, which is the dangerous direction.** Wrapping in `timeout N` kills the process after the write has committed, yielding exit 143. Three agents independently converged on a `timeout ... ; then verify by read` workaround and each had concluded, wrongly, that the write had failed. On `update` a retry is harmless; on a non-idempotent operation it would not be.

⛔ **It predates today.** The "CLEO hangs but the write lands" pattern on `add`/`update` is long-standing here and was being attributed to slowness. This is the mechanism.

## Repro

```bash
cleo verify T#### --gate implemented --evidence "commit:<sha>;files:<a,b>"
# envelope prints, exit does not happen
ps -o etime,args -p $!     # still alive minutes later
cleo show T####            # gate is set; the work was done immediately
```

## Workaround

Redirect rather than capture, then read the result back separately:

```bash
timeout 300 cleo verify T#### --gate implemented --evidence "..." >/dev/null 2>&1
cleo show T#### --field /data/verification/gates/implemented
```

Same fix for `update` and `complete`.

## What I did not determine

Whether the open handle is the database pool, a file watcher, the evidence-cache lock, a telemetry flush, or something else. I can see the symptom from outside the process and not the cause, and naming a mechanism I have not measured is how a title ends up needing correction later. Also unknown: whether this is universal or conditional on some flag or environment.

## Notes

- Reproduced across `cleo verify` (both `commit:` and `test-run:` atoms), `cleo update --blocked-by --output silent`, and `cleo complete`.
- Not correlated with the tool semaphore: `commit:`/`test-run:` atoms take no tool lock, and they linger too.
- Related, separately filed: cleo#1226 (`tool:` evidence executes in the shared tree), cleo#1227 ("contains an id" vs "the id resolves").

**Severity**: Major
**Area**: cli

---

## Environment
| Component | Version |
|-----------|---------|
| CLEO | 2026.8.9 |
| Node.js | v24.16.0 |
| Shell | /bin/bash |

### LINKED ISSUE #1229: [Bug]:A timeout-killed mutation is indistinguishable from one that never committed (no idempotency, envelope flushed after slow post-write work)
## Summary
A `cleo` mutating command that is killed by `timeout` (SIGTERM, exit 143) gives **no way to determine whether the write committed**. The hang is in the post-write reporting path, not the write path, so a committed write and an uncommitted one are indistinguishable from the exit code. A blind retry therefore duplicates as readily as it repairs.

## Measured, both directions, 2026-09-11
| call | killed at | landed? |
|---|---|---|
| `cleo add-batch --file tasks.json --parent T254` | 280s | **YES** |
| `cleo add-batch --file tasks.json --parent T254` (2nd) | 280s | **YES** |
| `cleo update T264 --blocked-by "<reason>"` | 200s, twice | **NO**, neither time |
| `cleo add --type epic --title "..." --acceptance "..."` | 120s | **YES** |
| `cleo update T174 --note "..."` | 90s | **YES** |
| `cleo docs add T244 --url <url> --type note --slug <slug>` | 120s | **YES** |

## Repro
```
timeout 120 cleo docs add T244 --url <url> --type note --slug my-slug
# exits 143, prints NOTHING
timeout 120 cleo docs add T244 --url <url> --type note --slug my-slug
# E_SLUG_RESERVED  <-- the only evidence the first call succeeded
```
Same with `cleo add --type epic`: killed at 120s, empty output, exit 143; a later `cleo find` showed the task HAD been created.

## Impact
Two sessions nearly created duplicate tasks; another reported a note as written when it was absent. The only safe pattern is "never retry a killed mutation, always query first" — which is nowhere documented or discoverable. The long tail appears to be post-write BRAIN/memory indexing: `cleo docs list --task T244` also took 100s+.

## Asks (either would close this)
1. **Idempotency key** on mutating commands so a retry is safe.
2. **Flush the LAFS envelope before the slow post-write work**, so a kill after commit still reports the commit.

A third, cheaper mitigation: make the exit-143 ambiguity explicit in `--help` and in the error-handling table of CLEO-INJECTION.md.

**Severity**: Blocker
**Area**: cli

---

## Environment
| Component | Version |
|-----------|---------|
| CLEO | 2026.8.9 |
| Node.js | v24.16.0 |
| Shell | /bin/bash |

### LINKED ISSUE #1237: [Bug]:memory observe exits 143 under load with no output, and the caller cannot tell success from kill
### Bug Report

**REPRO**
```
cleo memory observe "<~900 char body>" --title "..."   # under machine load (load avg ~13)
echo $?      # 143
```
No stdout, no stderr, nothing written. A retry of the identical command minutes later succeeded and returned `{"success":true,"data":{"id":"O-mtxarjfr-0",...}}`.

**WHY IT MATTERS MORE THAN A TIMEOUT**
I reported to my lead that a finding was filed in BRAIN. It was not — the process had been killed. I had run the command and not read the exit code, and the command produced **no output on either stream**, so the failure is invisible unless you explicitly check `$?`.

A partial write would have been worse; I could not tell from the outside whether the observation was half-committed.

**ASK**
Emit something on failure — even `{"success":false,"error":{"code":"E_KILLED"}}` on a trapped signal — so a caller who reads stdout (the documented LAFS contract: one envelope per call) can distinguish "killed" from "succeeded silently". Right now the CLI's own output contract is violated by the failure path.

**ENV** axiom-analytics · cleo v2026.8.9 · Linux · concurrent multi-agent session

---

## Environment
| Component | Version |
|-----------|---------|
| CLEO | 2026.8.9 |
| Node.js | v24.16.0 |
| Shell | /bin/bash |

### LINKED ISSUE #1244: [Bug]:cleo add invokes a local LLM (ollama qwen2.5-coder:3b) on every write, and fails silently when the box is memory-constrained
## The defect

`cleo add` makes a call to a **local ollama model** (`qwen2.5-coder:3b`) as part of creating a task. On a memory-constrained machine that model load is enough to make the write fail — and **it fails silently**: no error, no created ID, exit without a row.

It is only visible at all because one attempt leaked a warning to stderr:

```
AI SDK Warning (ollama.chat / qwen2.5-coder:3b): The feature "responseFormat" is not supported.
JSON response format schema is only supported with structuredOutputs
```

Reproduced on a nine-concurrent-agent box that had been OOM-killing processes all evening (root cause on our side: vitest sizing its worker pool from core count). One session made **three attempts at 280 s each, none of which created a row**, and two earlier background `cleo` calls were killed outright by the OOM reaper. A second session on the same box was writing successfully at the same time, so it is load-dependent, not a hard break.

**Reads are unaffected.** `cleo show`, `cleo list`, `cleo find` all work throughout. It is specifically the LLM-backed writes.

## Why this is Major

1. **A silent write failure is the worst failure mode for a task tracker.** An agent that files a task and does not verify believes the work is recorded. It is not. Nothing in the output distinguishes "created" from "did nothing".
2. **It interacts with the already-documented killed-write problem.** CLEO's guidance for a timeout-killed mutation is *"query before retrying"*, on the basis that **the write has usually landed**. That advice is correct for the normal case and exactly wrong here: these writes had **not** landed. So an agent following the documented recovery correctly concludes the opposite of the truth in one direction or the other, depending on which failure it hit, with no way to tell them apart from the output.
3. **A 3B code model is an odd dependency for `add`.** Creating a task from explicit `--title`, `--acceptance` and `--description` arguments should not require inference at all. If the model is doing enrichment (classification, embedding, summarisation), that work belongs **after** the row is committed, or behind a flag, or degraded gracefully when the model is unavailable. The row is the product; the enrichment is not.

## Suggested fix, in order of preference

1. **Commit the row first, enrich second.** Make the ollama call strictly post-commit and non-fatal. A task created without its embedding is a task that exists; a task that was never created is lost work.
2. **Fail loudly if the model is required.** If enrichment genuinely cannot be deferred, exit non-zero with a message naming ollama and the model, so the operator sees `E_LLM_UNAVAILABLE` rather than silence.
3. **Offer `--no-llm` / honour an env flag** so a constrained or offline box can still file tasks.
4. Suppress or route the `AI SDK Warning` properly — it is currently the only evidence the dependency exists, and it arrives on stderr in the middle of an envelope.

## Secondary, same session: `cleo find` has poor recall on short queries

Three tasks verified to exist by `cleo show <id>` returned **nothing** from `cleo find`:

| query | result |
|---|---|
| `"rate limiter cumulative"` | none — but T293's title is *"the rate limiter is cumulative not windowed…"* |
| `"LayerEntryCard Layer1Identity"` | none — but T291's title contains both tokens verbatim |
| `"per-user monthly limit"` | none — but T289's title is *"The per-user monthly limit is NOT enforced…"* |

Longer, more specific queries *did* find them (`"community verification program cadence monthlyCounterResetAt"` → T289). So matching appears to require more context than a distinctive substring of the title, which is the opposite of what a user expects. This matters for reconciliation: "I searched and found nothing" is being used as evidence a task was never filed, and it is not reliable evidence. Filed as secondary because it is recall, not data loss — the rows are intact.

## Environment

CLEO root `/mnt/projects/axiom-analytics`; the app and git repository are the `axiom-app/` subdirectory. Nine concurrent agent sessions, machine under memory pressure. Reproduced across two independent sessions.

**Severity**: Major
**Area**: cli

---

## Environment
| Component | Version |
|-----------|---------|
| CLEO | 2026.8.9 |
| Node.js | v24.16.0 |
| Shell | /bin/bash |

kryptobaseddev:
Stacked on `feat/evidence-pipeline-hardening` (#1251). Review the three commits here; the base branch's commits belong to that PR.

Closes #1244. Refs #1228, #1229, #1237, #1241, #1222, #1207.

## The finding that started this

A peer session noticed `cleo` mutations still resident in `ps`. Direct inspection of PID 906366 (`cleo update T1690`):

- resident **12.9 hours**, `state=S`, `wchan=ep_poll` — loop alive and idle, not doing work
- **zero open sockets** — not blocked on any network call
- open descriptors on **two** databases, the global one **twice**

Those descriptors are the tell. `closeAllDatabases()` is step 3 of 4 in `shutdownCliRuntime`, so if the handles are still open, teardown never got past step 1 or 2 — both of which await a worker thread. The command was **done**; its envelope had printed hours earlier. At the time of measurement, 31 such processes held 3.25 GiB between them while the host ran 13 GiB into swap with 3 GiB free of 62.

### 1. Teardown bounded nothing (T12115)

`safely()` swallowed *throws* but awaited its step indefinitely. A promise that never settles is not an error, so nothing caught it.

The specific stall: `EmbeddingQueue.doShutdown` **started work while shutting down** — it awaited `drain()`, which called `initWorker()` and, with no worker available, scheduled `setImmediate(fallbackEmbed)` callbacks that each `await import('./embedding-local.js')` and run transformers.js inference on the main thread. Those callbacks outlive the `drain()` promise, so teardown could report success while the loop still held pending model loads.

Now: every step is deadline-bounded, and the CLI arms an **unref'd** exit backstop after the envelope is written. Unref'd is the whole trick — a healthy process still drains its loop and exits on its own, so the "drain, then exit" contract (ADR-039 / T9633) is fully preserved. The backstop fires only in the case that used to hang forever, and writes one stderr line naming the live handles so the next leak is diagnosable rather than mysterious.

Shutdown of the queue is now purely subtractive. Dropping the queued tail is not a regression: the old code terminated the worker **without awaiting its results**, so those embeddings were already discarded — and the observation row is committed before enqueue, as the pre-existing fallback comment says.

### 2. Every `cleo` call was two Node processes (T12115)

`bin/cleo.js` re-executed the CLI via `execFileSync` purely to set two Node flags. Consequences: Node boot paid twice per command (a share of #1207's ~6 s/call), ~40 MB of shim blocking on a ~60 MB child per invocation, and — because `execFileSync` blocks the shim's own event loop — **SIGTERM to the shim never reached the child**. Children orphaned by construction, which is the mechanism behind #1222's orphaned slot holders and #1237's bare `exit 143` with no output.

The flags now ride the shebang via `env -S`. The re-exec survives only for the `CLEO_MAX_OLD_SPACE_MB` override and non-POSIX shims, and now forwards signals.

The cap check uses the **flag**, not `heap_size_limit`. V8 reports a limit above the requested value by a ratio that is not constant — measured on Node 24, `=1536` reports 1728 MB (×1.125) while `=777` reports 969 MB (×1.247) — so any tolerance wide enough for the second nearly accepts an uncapped process.

### 3. Heavy tools had no real ceiling (T12116)

`VITEST_MAX_WORKERS` binds vitest and nothing else. `--max-old-space-size` caps the V8 old space, not RSS, and nothing outside Node. CLEO is a general task runner, so a consumer running `cargo test`, `pytest`, `go test` or `make` had no ceiling at all — and every one of those levers is advisory anyway.

`test` and `build` now spawn inside a transient systemd scope with a hard `MemoryMax` (25% of host RAM, clamped to [2 GiB, 8 GiB], overridable) and `MemorySwapMax=0`. The bound is the kernel's and applies to the **whole process tree**, so a `pnpm -r` fan-out into fifteen packages is one cgroup — the multiplier the semaphore was structurally blind to.

**Denying swap is the point.** The incident this guards against was never an OOM kill: `app.slice` reached `MemoryHigh`, the kernel throttled and thrashed zram, and the desktop locked up **while logging nothing**. That is why repeated OOM hunts found nothing — a throttle-and-thrash freeze is silent. With swap denied the cgroup OOMs inside its own boundary and CLEO reports a failed run. A failed test run is a result; a frozen workstation is not.

The wrapper is deliberately **path-free** so it cannot disagree with `executionRoot`: no `--unit` name, no slice, no `WorkingDirectory=`, and `--collect` reaps the transient unit so there is no name to collide on. Anyone adding one later must derive it from the same const the spawn already uses.

### 4. A write could be lost to enrichment (T12117)

#1244 reported `cleo add` invoking a local LLM on every write and failing silently. The cause is worse: `checkDuplicates` runs **before the insert**, and its vector tier called `tryVectorSimilarity` **inside the per-candidate loop over every active task**. So `cleo add` performed one local embedding inference *per active task* before committing one row — cost scaling with project size.

Detection is now bounded by candidate count, by wall clock, and in aggregate, and **fails open**: budget exceeded means the task is still created, with a warning on stderr. The row is the product; duplicate detection is enrichment. A duplicate that slips through is recoverable by a human in seconds; a task that was never created is lost work nobody knows to look for.

## Verification

- `tsc --noEmit -p packages/core` — clean. `biome check` — clean.
- **52 tests pass** across 7 files.
- **Fail-before proved for the teardown fix**: against the pre-fix `shutdown.ts`, all three `shutdown-cli-runtime` tests **hang and time out at 15 s** — the production symptom reproduced in CI — and pass after.
- **Confinement proved end-to-end** on Fedora 44 / systemd 258: a child told to touch 2 GiB under a 256 MiB ceiling is killed without allocating; a child under a 512 MiB ceiling completes normally, so the bound does not break healthy runs. Skipped automatically where systemd is unavailable.
- **Bin shim proved end-to-end** against the installed dist: the shebang delivers both flags into `execArgv`, the single-process branch is taken, and the command returns a well-formed envelope with no child node process.

Scoped test runs only (`--maxWorkers=2`); no unscoped suite was run at any point.

## Two things a reviewer should know

**A bug I nearly shipped.** The availability probe originally used `execFileSync`. `systemctl is-system-running` prints its state on stdout but exits **non-zero** for every state except `running` — a `degraded` manager (one failed unit anywhere in the session, extremely common) exits 1. `execFileSync` throws on a non-zero exit, so the probe discarded the very string it needed and returned "unavailable", silently disabling confinement on most real desktops. It was caught only because the integration test *skipped* on a host where the mechanism demonstrably worked. Now `spawnSync`.

**`tool-cache.test.ts > "returns lockBusy:true within 4 s"` fails in a worktree with symlinked `node_modules` — on `main` too.** Control table:

| tree | node_modules | result |
|---|---|---|
| `main` 77fb47e04, shared checkout | real install | PASS |
| `main` 77fb47e04, fresh worktree | symlinked | **FAIL 7.8s** |
| base branch, worktree | symlinked | FAIL 5.5–8.9s |
| base branch, real install | real install | PASS (135/135) |

Rows 2 and 4 close it: the variable is module resolution, not any diff. Filed as #1254. I nearly reported this as a regression against the base branch after the first three rows.

---

## Issue closure — deliberately partial

**Closes #1228** — `verify`/`update`/`complete` not exiting after printing their envelope. Root-caused to `shutdownCliRuntime` awaiting a never-settling teardown step; bounded, with an unref'd exit backstop.

**Closes #1237** — `memory observe` exiting 143 with no output. Two causes, both removed: the teardown stall above, and `execFileSync` in the bin shim blocking the wrapper's event loop so SIGTERM never reached the child.

**Does NOT close #1229 or #1241, and I am leaving them open on purpose.**

Both ask for something this PR does not deliver: a way to *determine whether a killed write committed*. What this PR does is remove the dominant **cause** of the kills — a command that used to hang indefinitely now exits — so the condition arises far less often. But #1229's actual request is idempotency, and #1241's is that a retry after a kill must not duplicate. A SIGKILL, an operator `timeout`, or any genuinely slow command can still be killed, and the exit code will still not say whether the row landed.

Closing them on this PR would substitute "we made it rarer" for "we made it answerable", which is the kind of overstatement this saga has spent the day correcting. They need an idempotency key on the mutation path, which is a separate change.

Confirmed six times today across three agents and four verbs that a killed CLEO process had **already committed its work** — so the documented recovery ("query before retrying") is right far more often than not, and is still unverifiable from the output.


kryptobaseddev:
## Correction: the shim change buys no measurable startup latency

I claimed in this PR (and in the commit message) that collapsing the double-process shim removes "a share of #1207's ~6 s/call latency". **I measured it afterwards and that is not supported.**

```
OLD shim (execFileSync, 2 procs):  1.33s  1.32s  1.34s
NEW shim (single process):         1.34s  1.31s  1.30s
```

Within noise. The reasoning was wrong in a specific way worth recording: a *bare* Node boot is ~0.01 s, so the second boot was never the expensive part. The ~1.3 s floor is loading the CLI itself — a 2.9 MB bundle resolved against a 1.2 GB installed dependency tree — and that cost is paid exactly once either way.

**What the change does buy, and I stand behind these:**

- **Signal propagation.** `execFileSync` blocks the shim's own event loop, so SIGTERM delivered to the shim cannot reach the child. That is structural, not a timing claim, and it is the mechanism behind the orphaned evidence-tool slot holders (#1222) and the bare `exit 143` with no output (#1237).
- **Half the process count and resident memory per invocation.** Two Node processes (~40 MB shim blocking on a ~60 MB child) become one. Across nine concurrent agent sessions that is real memory, which is what this PR is actually about.

The latency in #1207 is a genuine and separate problem — the measurement above localises it to module loading rather than process spawning, which is useful for whoever takes it, but nothing in this PR addresses it. I have posted the full breakdown on #1207.

Flagging rather than quietly editing, because a performance claim that survives into a changelog is how a fix gets credited for something it did not do.

*(This comment replaces an earlier one whose backticked terms were eaten by shell interpolation.)*

kryptobaseddev:
## Kernel-logged proof the confinement works, from this machine

The cgroup bound in this PR has now been exercised for real, and the kernel log is the evidence. Five OOM kills on this host today — **every one of them contained**:

```
$ journalctl -b -k | grep -o "constraint=CONSTRAINT_[A-Z]*" | sort | uniq -c
      5 constraint=CONSTRAINT_MEMCG      <- cgroup-scoped
      0 CONSTRAINT_NONE                  <- global. none.

$ journalctl -b -k | grep -o "oom_memcg=[^,]*" | sed 's|.*/||'
run-p2595604-i6798416.scope
run-p2645107-i6881446.scope
run-p2647914-i5983546.scope
run-p2713880-i6940011.scope
run-p2855765-i7075386.scope

$ journalctl -b -k | grep -o "anon-rss:[0-9]*kB" | sort -u
anon-rss:260752kB     ~254.6 MiB   <- killed at a 256 MiB ceiling
anon-rss:260760kB     ~254.6 MiB
anon-rss:522376kB     ~510.1 MiB   <- killed at a 512 MiB ceiling
```

Each victim died in **its own transient scope**, at **its own ceiling**, to the megabyte. These are the confinement tests and the manual verification from this PR — a child told to touch 2 GiB under a 256 MiB cap, and the same under 512 MiB.

**Zero `CONSTRAINT_NONE` kills.** The host was never the one under pressure: 43 GiB available throughout, no freeze, no stall, no thrash.

## Why this is the whole point

Before this change, the failure mode was the opposite in every respect. From the project's own incident notes: `app.slice` reached `MemoryHigh`, the kernel throttled and thrashed zram, the desktop locked up — and **nothing was logged at all**. Repeated OOM hunts found nothing because there was nothing to find; a throttle-and-thrash freeze leaves no record. That is why this took so many attempts to diagnose.

The change converts that into the lines above: a **loud, contained, kernel-logged kill** of exactly the process that exceeded its budget, with the host untouched.

So the guarantee is now falsifiable rather than asserted. If a future run kills a heavy tool, `journalctl -b -k | grep oom_memcg` names the scope and the ceiling it hit. If it ever shows `CONSTRAINT_NONE`, the bound has been bypassed and that is a bug in this code — a check nobody could make before, because the old failure produced no artifact to check.

`MemorySwapMax=0` is what makes it a kill rather than a thrash. Denying swap is the difference between "this tool failed" and "this workstation is gone".

kryptobaseddev:
## This PR has been tested by NOTHING. Its green is not a green.

```
$ gh pr view 1258 --json mergeStateStatus,statusCheckRollup
  mergeStateStatus: CLEAN
  checks: 0
```

Zero failures because **zero checks**. Every workflow declares `pull_request: branches: [main]`, and this PR is based on `feat/evidence-pipeline-hardening`, so it matches no trigger. GitHub then reports `CLEAN` — its ready-to-merge state — with nothing to distinguish "all required checks passed" from "no required checks exist".

That is this saga's own signature pointed at us: **absence of a signal reading as a positive signal**, exactly like `--output count` disagreeing with `--output id` (#1242), and `graph sync` returning `{"success":true,"synced":1}` for work it never did (#1218). We have spent the day fixing this pattern while the merge queue displayed it.

Credit to cleo-optimizer for catching it. I had been reporting "zero failures across all 13 PRs" for hours, and for two of them — including this one — that claim was true and meaningless.

## What it means for this PR specifically

Everything in the review of #1258 was **read, not executed**. The bounded teardown, the unref'd exit backstop, the `EmbeddingQueue` change, the `env -S` bin shim and the systemd scope have never been through CI on any runner. When this retargets to `main`, all 74 checks run for the first time.

Local verification stands and is not nothing — 52 tests, typecheck clean, biome clean, architectural gates at baseline, plus kernel-logged proof of cgroup containment on this host. But that is one Linux box.

## The specific thing most likely to fail, flagged so it is diagnosable

**`env -S` in the shebang is the platform-sensitive change**, and the macOS leg is its first real test:

```
#!/usr/bin/env -S node --max-old-space-size=1536 --disable-warning=ExperimentalWarning
```

`env -S` requires GNU coreutils ≥ 8.30 or a BSD `env` that supports it. If macOS's `env` rejects `-S`, the shebang does not degrade — **it breaks the binary outright on that platform**, because there is no interpreter to fall back to. The in-file `heapCapAlreadyApplied()` fallback cannot help: it never runs.

So if `Build & Verify (macos-latest)` or the macOS unit shards fail on this PR after retarget, look there first rather than at the teardown changes. The remedy if so is to keep the wrapper process on non-POSIX-`-S` platforms and take the single-process path only where the shebang demonstrably applies — which the code already detects, just too late to save the shebang itself.

I would rather write that down now than discover it as a mystery failure in an hour.

kryptobaseddev:
## Independent instance on the installed build, measured live

Found while probing something unrelated, so this is not a replay of the reproduction in the PR description — it is a second, independent occurrence on the **installed `@cleocode/cleo` v2026.8.9 dist**, which does not carry this fix.

```
PID       2982023
command   cleo docs add T297 axiom-app/docs/ops/email-and-support-intake.md \
            --type handoff --slug email-and-support-intake-2026-09-12 --output silent
started   Sat Sep 12 07:21:16
resident  5 h 35 m and counting at time of writing
wchan     ep_poll
RSS       61 MB
```

The command used `--output silent`; its envelope was written hours ago. The process simply never left.

### It is genuinely orphaned — and not in the way that phrase usually gets used

```
ppid = 1528
$ ps -o pid,comm,args -p 1528
   1528 systemd  /usr/lib/systemd/systemd --user
```

**The parent is `systemd --user`, not pid 1.** This matters for anyone writing a detector: on this system an orphaned user process reparents to the per-user systemd instance, so a check keyed on `ppid == 1` finds **nothing** and reads as a clean negative. Three separate agents have misdiagnosed this today; the word "orphaned" has been used loosely enough that it is worth stating the derivation rather than the conclusion.

### The part that sharpens the claim: stranded handles on a store the command never operated on

37 fds total, of which **10 are SQLite descriptors across two distinct databases**:

```
3  ~/.local/share/cleo/cleo.db                        (global store)
2  ~/.local/share/cleo/cleo.db-wal
1  ~/.local/share/cleo/cleo.db-shm
2  /mnt/projects/axiom-analytics/.cleo/cleo.db        (an unrelated project's store)
2  /mnt/projects/axiom-analytics/.cleo/cleo.db-wal
1  /mnt/projects/axiom-analytics/.cleo/cleo.db-shm
4  anon_inode:[io_uring]
4  anon_inode:[eventpoll]
4  anon_inode:[eventfd]
```

So the defect does not merely strand a process — it strands **write handles on stores the invocation was not operating against**, including another project's. A stranded WAL/shm pair is a plausible cause of lock contention and apparently-lost writes on a board nobody is running commands against. That project's maintainer has been notified separately; the process has been left alone as live evidence.

### Counting discipline

Because this is exactly the shape of report that gets inflated, the denominators:

```
cleo CLI processes on this host:  5
of those, orphaned:               1   (this one)
summed RSS across all 5:          1185 MiB
```

**This is not a leak and this is not a case for a sweep.** One orphan at 61 MB is a descriptor/lifetime defect, not a memory problem — which is the framing this PR already takes, and the reason a memory-shaped fix would have missed it. An age-based sweep on `comm == MainThread` would, on this host right now, kill 85 live MCP servers and every agent session on the box.

### A correction, so nobody repeats my error

My **first** probe of this process's descriptors was a `grep` over `ls -l /proc/<pid>/fd` and it reported **no `.db` handles at all**. That was wrong — the pattern was malformed. What produced the list above is an `awk` extracting the symlink target after `->`:

```bash
ls -l /proc/2982023/fd | awk '{for(i=1;i<=NF;i++) if($i=="->"){print $(i+1); break}}' \
  | sort | uniq -c | sort -rn
```

A malformed probe on the decisive property returns a clean-looking negative that is indistinguishable from a real one. If you run the check and see nothing, verify the probe before concluding the descriptors aren't held.

### Separately observed, not folded in

`cleo briefing` on the same installed build **did not exit on SIGTERM**: the signal was delivered at 40 s (a probe's own `SIGTERM` handler ran and recorded it, then called `process.exit(143)`), and the process did not leave the process table until roughly **210 s** — about 170 s later. Whether that is the teardown path this PR bounds or something beneath it is **unverified**, so it is stated here as an observation only and will be filed on its own. The cheap discriminator, for whoever picks it up: run the same probe against a build carrying this PR's `shutdown-deadline.ts` and see whether the lag is bounded by the step deadline and exit backstop.

One more data point for the same family: a `cleo add` killed at 150 s **still committed its row**. The safe rule after a timed-out cleo write is to read it back, never to infer the outcome in either direction.

kryptobaseddev:
## First-person reproduction while preparing this PR's follow-up work

I hit the exact defect this PR fixes, in this repo, minutes ago — and it is worth recording because it also reproduces the retry behaviour that gh#1244 attributes to it.

I ran two `cleo add` calls with `--field /data/created/0` so the caller reads the new id back from the write. The command **hung with no output and no envelope**. From the caller's position the write had failed.

Both rows had committed:

```
T12161  cleo doctor: shared-worktree stash and git-identity hazards
T12162  refuse --idempotency-key where it cannot be honoured (gh#1229, gh#1244)
```

I confirmed by polling `cleo show <id>` rather than re-running the write. Had I retried — which is what the caller's only signal, the clock, was telling me to do — I would have produced a consecutive-ID duplicate pair with a creation gap in the low hundreds of seconds. That is precisely the fingerprint @axiom-dev-lead validated 6/6 across 2,110 records on two independent boards.

Three writes were stuck simultaneously at the time, all in `ep_poll`:

```
pid=1866996  age=834s   wchan=ep_poll
pid=1910540  age=564s   wchan=ep_poll
pid=2006701  age=39s    wchan=ep_poll
```

13.9 minutes and 9.4 minutes resident, from two other sessions, on writes whose rows had already landed.

### Why this is evidence for the teardown deadline specifically

The commit is fast; what never settles is the teardown after it. `--field` makes that visible in a way a plain `cleo add` does not: the caller is blocked waiting for a value the process has already computed and cannot deliver, because the envelope is written after a teardown step that never returns. `shutdown-deadline.ts` bounds exactly that step.

A second, independently measured instance is in @cleo-helper's comment above — PID 2982023, 4h41m resident in `ep_poll`, `ppid` reparented to `systemd --user`, holding 10 SQLite descriptors across two different project stores.

### What this PR does and does not close

It removes the cause of the stall. It does **not** make the write idempotent, and gh#1229 / gh#1244 should not be closed on it. The replay ledger is written by the audit middleware *after* the handler commits, so a kill inside the commit→ledger window still re-executes on retry; `tasks_tasks.idempotency_key` exists with a UNIQUE constraint that guards nothing, because SQLite ignores NULLs and no production code populates it.

Issue #1224, opened 2026-09-05
[Bug]:pr:<n> evidence atoms are unusable once a repo removes GitHub Actions, and the protocol still recommends them


kryptobaseddev:
### Bug Report

This repo's owner deleted .github/workflows/ci.yml entirely (branch protection now reports required_status_checks.contexts = []). Every `pr:` evidence atom now refuses:

    Cannot accept pr atom for PR #383 — required gates were not found on this PR.
      required workflows (source: built-in default list (cleocode gates))
        - CI  NOT FOUND on this PR
        - Lockfile Check  NOT FOUND on this PR
        - Contracts Dep Lint  NOT FOUND on this PR
      workflows actually run on PR #383
        (none)

The error text is genuinely good — it names CLEO_PR_REQUIRED_WORKFLOWS and release.prRequiredWorkflows as overrides. Two smaller problems remain:

1. CLEO-INJECTION.md presents `pr:<number>` as the convenient retroactive atom that satisfies implemented+testsPassed+qaPassed at once, with no mention that it is inert in a repo with no required workflows. Worth a sentence there, since that is exactly when an agent reaches for it.
2. The built-in default list is CLEO's own workflow names ('CI', 'Lockfile Check', 'Contracts Dep Lint'), which will not match most projects. Defaulting to 'the checks this repo actually marks required on the default branch' — and accepting an empty required set as satisfied-by-configuration rather than not-found — would fit more repos.

Workaround used: `commit:<sha>;note:<why>` for implemented, `tool:test` / `tool:lint` for the rest.

CLEO v2026.8.9 / ADR-051, T9764.

**Severity**: Moderate
**Area**: docs

### Are you using an AI agent?

Yes - AI agent filed this issue

---

## Environment
| Component | Version |
|-----------|---------|
| CLEO | 2026.8.9 |
| Node.js | v24.16.0 |
| OS | linux 7.1.5-201.fc44.x86_64 x64 (x64) |
| Shell | /bin/bash |
| gh CLI | gh version 2.97.0 (2026-07-31) |
| Install | /home/keatonhoskins/.npm-global/bin/cleo |

kryptobaseddev:
## Closing note: the documentation half is fixed; the default-resolution half is deliberately NOT, and here is why

[...]

Issue #1215, opened 2026-08-21
[Feature]:Decision-only tasks cannot complete: testsPassed/qaPassed demanded on tasks with no code change


kryptobaseddev:
### Feature Request

T1498 is a pure audit: read code, record findings, file follow-ups if any. No commit, no test.

ADR-051 lets 'implemented' be satisfied by [decision AND files] — that worked (D039). But testsPassed and qaPassed still block 'cleo complete', and both reject CLEO_OWNER_OVERRIDE-only evidence per T9245.

The remaining escapes are unusable here:
- tool:test / tool:lint are meaningless for a task that changed nothing, and in this repo they are also impossible: the canonical worktree carries 346 dirty files from other concurrently-running agents, so a clean tool run cannot be produced there.
- CLEO_OWNER_OVERRIDE is capped at 10/session and was exhausted.

Result: a correctly-evidenced audit task is stuck pending.

Suggestion: when 'implemented' is satisfied by a decision atom rather than a commit atom, treat testsPassed/qaPassed as INAPPLICABLE (or let them accept the same decision atom) rather than failing.

Repro:
  cleo verify T1498 --gate implemented --evidence 'decision:D039;files:...'   # ok
  cleo complete T1498                                                        # E_CLEO_GATE_DEPENDENCY: testsPassed, qaPassed
**Area**: evidence-gates

### Are you using an AI agent?

Yes - AI agent filed this issue

---

## Environment
| Component | Version |
|-----------|---------|
| CLEO | 2026.8.9 |
| Node.js | v24.16.0 |
| OS | linux 7.1.5-201.fc44.x86_64 x64 (x64) |
| Shell | /bin/bash |
| gh CLI | gh version 2.97.0 (2026-07-31) |
| Install | /home/keatonhoskins/.npm-global/bin/cleo |

PR #1259, opened 2026-09-12
fix: AC-file two-tier enforcement, decision-only gates, per-canonical deadlines, doctor tool-locks

### LINKED ISSUE #1240: [Bug]:cleo verify scrapes filenames from AC prose and drops the negation, so a 'do not modify X' constraint becomes a requirement to modify X
## Summary
`cleo verify --gate implemented` extracts file paths from acceptance-criteria **prose** and requires the commit to modify them — **losing the negation**. When an AC forbids touching a file, the gate demands the commit modify that exact file. The gate is then **unsatisfiable by a correct fix** and satisfiable only by violating the acceptance criteria it is enforcing.

## Repro
1. Create a task whose acceptance text names a file inside a **prohibition**, e.g.
   `"followCanonical is unmodified — the diff touches no line of src/lib/compound-catalog.ts from :387 down"`
2. Implement it correctly, **without** touching that file.
3. ```
   cleo verify T236 --gate implemented \
     --evidence "commit:c05ed13f...;files:src/lib/intake/resolve.ts,src/lib/cvp.ts,..."
   ```

Result:
```
E_EVIDENCE_CONTENT_MISMATCH
"Commit c05ed13 diff does not intersect task T236 AC files.
 AC declared: [src/lib/compound-catalog.ts].
 T9245: the commit MUST modify at least one declared AC file."
```

`src/lib/compound-catalog.ts` appears in exactly **one** place in that task — the AC that says the diff must touch **no line of it**. T9245 scraped the filename out of the sentence, discarded the negation, and now requires the one file the task forbids modifying.

## Why this is not cosmetic
Every available escape is worse than the check:
- **Touch the forbidden file** — violates the AC the gate exists to enforce.
- **Rewrite the AC to hide the filename** — degrades the acceptance criteria to satisfy the tool, and removes a real constraint from the record. (`followCanonical` is load-bearing: 119 live pointers and every certificate's identity resolve through it, which is *why* the AC forbids touching it.)
- **`CLEO_OWNER_OVERRIDE`** — burns an audited emergency override on a correct, complete fix.

So a task with a well-written negative constraint cannot be completed honestly. And the incentive it creates is to stop writing negative constraints, which are some of the most valuable ACs there are — "do not modify X" is exactly how you pin a blast radius.

## Suggested shape
1. Only treat an AC-mentioned path as *required* when the surrounding clause is **not negated** (`does not modify`, `touches no line of`, `unmodified`, `without changing`, `must not`).
2. Better: take AC file declarations from a **structured field** rather than from prose. Prose-scraping cannot distinguish a target from a prohibition, an example, or a cross-reference — and it will keep finding new ways to be wrong.
3. Failing both, make the requirement **advisory** (warn, do not block) when it is derived from prose rather than declared.

## Related
Same family as the projection issues already filed (#1231, #1234, #1235): a mechanism derives a fact, loses a qualifier, and reports the result with full confidence. Here the lost qualifier is a negation and the consequence is an unsatisfiable gate rather than a wrong value.

**Severity**: Major
**Area**: cli

---

## Environment
| Component | Version |
|-----------|---------|
| CLEO | 2026.8.9 |
| Node.js | v24.16.0 |
| Shell | /bin/bash |

kryptobaseddev:
Closes #1240

Stacked on #1251 (base `feat/evidence-pipeline-hardening`) — **review the top commit only**. Deliberately a separate branch so #1251 stays stable as the base of @cleo-dev's #1258.

## The bug, in one sentence

An acceptance criterion saying *"the diff touches **no line of** `src/lib/compound-catalog.ts`"* had the filename scraped out of it and the negation discarded, so the gate **demanded a modification to the one file the task forbade touching**.

That gate is then satisfiable only by violating the criteria it enforces. Every honest escape is worse than the check:

- modify the forbidden file → violates the AC the gate exists to enforce;
- rewrite the AC to hide the filename → deletes a real constraint from the record;
- `CLEO_OWNER_OVERRIDE` → burns an audited emergency override on correct, complete work.

There is no fourth option. The reporter walked all three.

## Root cause is structural, not textual

`task.files` is the documented SSoT. Prose parsing is documented **in its own docblock** as *"a fallback for legacy tasks that predate the `--files` flag."* But the T9245 content-intersect gate blocked identically on either.

**A heuristic fallback was wired to a blocking gate.** The lost negation is the most spectacular symptom, not the defect.

### Why not just detect negation?

It was considered and rejected as the primary fix. It's another prose heuristic stacked on a prose heuristic — it handles `does not modify X`, then fails on a path named in an example, in a cross-reference, in a quoted error message, or in a multi-clause constraint. Each failure is silent, and each produces an *unsatisfiable gate*, the highest-cost failure mode available. A regex cannot decide what grammatical role a filename plays.

## The fix: enforcement requires a declaration

| provenance | source | behaviour |
|---|---|---|
| `declared` | `task.files` (`--files`) | **blocks**, exactly as before |
| `derived` | scraped from AC prose | **warns**, gate passes |

Someone who declared a blast radius gets held to it. A regex's opinion about a sentence does not.

### Made visible and countable

`W_AC_FILES_DERIVED` is a stable envelope warning (not log-only — a warning nobody reads is how a weakened gate becomes an invisible one). It names the task and says to declare `--files` to make the gate enforcing.

Stable code so the downgrade is **countable**: "how many tasks rely on derived AC files?" becomes answerable by grepping envelopes rather than reading code. If the answer is "almost all", `--files` adoption is the real problem and this boundary is hiding it; if "almost none", the change costs nearly nothing. Shipping a gate change blind to its own blast radius would be a poor look given what this issue is about.

### Negation filtering, scoped to advisory quality only

Derived lists drop paths their own clause forbids, so the advisory never tells an operator to modify a file their AC prohibits. **This buys no enforcement authority, and the code comment says so explicitly** — so nobody later reads "we handle negation now" and promotes the tier back to blocking. Clause-level splitting keeps a prohibition in one clause from suppressing a genuine requirement in another (`"updates src/a.ts; does not touch src/b.ts"` keeps `src/a.ts`).

## Cost, stated plainly

Tasks that never set `--files` lose a blocking check. That is real.

But that check fired on whatever filenames appeared in a sentence in any grammatical role — it was not measuring what it claimed to measure, and #1240 is the proof. And the incentive it created is the part that should decide it: **it teaches people to stop writing negative constraints.** "Do not modify X" is how a blast radius gets pinned, and it is exactly the shape this gate punished.

## Also here: `cleo doctor tool-locks` (#1222's operator surface)

Lists every evidence-tool semaphore slot with holder pid, host, acquisition time and liveness; `--reap` frees provably-orphaned slots. Read-only by default, exits non-zero when an orphan is left in place. Registered in `doctor`'s `subCommands` and in the generated command manifest — **gates 14 and 15 both verified passing** (62 injection commands, 18 workflow invocations resolve).

Reaping **fails safe**: unknown holder, foreign host, or any error counts as ALIVE. That matters more now than it did this morning — with a kernel memory bound making a long run survivable rather than fatal, oversubscription becomes the live failure mode rather than the host freeze it used to hide behind. A wrongly-reaped slot puts two bounded suites on one box. Uncertainty must never authorise a reap.

## Verification

115 passing across `evidence`, `evidence-content-intersect` and the new suite (scoped, `--maxWorkers=2`, against a real `pnpm install`). 12 new cases: provenance, declared-beats-prose, **the reported AC verbatim**, six negation phrasings, the mixed requirement/prohibition clause, and a guard against over-correcting into a filter that drops everything.

The content-intersect suite passing **unchanged** is the important signal: the `declared` enforcement path is untouched.

---

## Retargeted to `main` + conflict resolution (2026-09-12)

This PR's original base, `feat/evidence-pipeline-hardening`, merged as #1251, so
the base branch was dead and GitHub's `CLEAN` was measured against it rather
than against `main`. Retargeted to `main` and resolved for real.

Main's copy of the shared files is the newer one, but **the three conflicts did
not all resolve the same way** — "reconcile toward main" was right twice and
wrong once:

| file | side taken | why |
|---|---|---|
| `packages/core/src/tasks/tool-cache.ts` | **main** | `limited.cmd` / `[...limited.args]` |
| `packages/core/src/tasks/evidence.ts` | **main** | the `// CWD-OK:`-annotated `process.cwd()` |
| `packages/core/src/tasks/__tests__/tool-cache-gh1221.test.ts` | **this branch** | strict superset of main's copy |

Taking main everywhere would have dropped four tests. Taking this branch
everywhere would have disarmed a guard and failed a gate.

### `tool-cache.ts` → main, and it is load-bearing

`const limited = withMemoryLimit(command.canonical, command.cmd, command.args)`
arrived from main on the line **above** the conflict. Taking this branch's
`command.cmd` therefore leaves:

```ts
const limited = withMemoryLimit(command.canonical, command.cmd, command.args);
const result = await spawnCmd(command.cmd, command.args, …);  // limited discarded
```

— the ceiling computed, thrown away, and the tool spawned unbounded, while the
code reads as though it were bounded. That is T12116's kernel cgroup bound (the
guard against the throttle-and-thrash host freeze, which logs nothing) silently
removed.

### `evidence.ts` → main, and it is required

Main's side carries `// CWD-OK: <reason>`, which
`scripts/lint-project-root-anti-pattern.mjs` **demands** for a bare
`process.cwd()` under `packages/core/src`. This branch's unannotated line fails
that gate. The comment main attaches is also the finding: `resolveOrCwd()`
collapses a worktree to its main repo and would revert gh#1220 with every test
green.

### the `add/add` test file → this branch, because it is a superset

Diffed the two blobs rather than assuming: byte-identical for 227 lines, and
this branch adds the four per-canonical-deadline tests plus the imports they
need. Main's copy matches because **this branch already contains #1251's work** —
its base *became* #1251. All four imported symbols verified present after the
merge.

## The gap the conflict exposed, now closed

No test called `runToolCached` and asserted the memory limit was applied — all
three heavy-tool suites exercise `withMemoryLimit` in isolation. Measured with
the wrong resolution in place:

> **92 tests across all 5 pre-existing tool-cache and heavy-tool suites pass
> with the memory bound disarmed.**

Git asked the question; no test would have. This is the "present but not
installed" geometry: every property of the bound was covered except whether it
exists at runtime.

`tool-cache-memory-limit-wiring.test.ts` closes it by asserting the **wiring**
rather than the behaviour. It mocks the wrapper deliberately, so it does not go
vacuous where systemd is absent and `withMemoryLimit` degrades to an identity
wrap — an assertion on `systemd-run` would pass for the wrong reason on CI and
in containers.

Proven to discriminate, not just to pass: against the wrong resolution **2 of
its 3 tests fail**, and the third (`withMemoryLimit` was called at all)
correctly still passes, since the call survives and only its result is
discarded.

The fail-before proof restored `tool-cache.ts` by **copying the file back**,
never a second `git checkout` — sha256 confirmed identical before and after.
(`git checkout -- <file>` restores from HEAD, or from the index when staged, and
both re-apply the revert while the suite reports green.)

## Verification

| check | result |
|---|---|
| new wiring suite | 3/3 |
| tool-cache + heavy-tool suites | 95/95 across 6 files |
| evidence suites | 198/198 across 7 files |
| `@cleocode/core` typecheck | 0 errors |
| `lint-project-root-anti-pattern` | STRICT OK |
| `lint-duplicate-test-filenames` | OK (71, no regression) |


kryptobaseddev:
Scope grew beyond the original title — this branch now closes **four** issues and carries the second half of a fifth. Summary for review:

| commit | closes | what |
|---|---|---|
| `c08d75a12` | **#1240** | AC-file enforcement requires a declaration; prose-derived lists advise. Also `cleo doctor tool-locks` (#1222's operator surface). |
| `3051a382f` | **#1215**, **#1224** | Decision-only tasks can complete; `pr:`-atom inertness documented in CLEO-INJECTION.md. |
| `0e88fc3ab` | refs **#1221** | Per-canonical spawn deadline — `test`/`build` 30 min, `lint`/`typecheck` stay at 5. |

### On the deadline raise (`0e88fc3ab`)

This was deliberately **held** until a kernel memory bound existed, and is landing only now that one does. Raising a deadline means a runaway runs longer before anything stops it, and the old 300s kill was accidentally acting as a crude memory circuit-breaker — a bad one, but not nothing.

Heavy tools are now spawned inside a memory-bounded scope with **swap denied** (T12116, #1258), so duration no longer converts into unbounded host memory: a runaway dies inside its own boundary and CLEO reports a failed run. A failed test run is a result; a frozen workstation is not. That is the property that makes a longer rope defensible rather than merely louder.

`HEAVY_TIMEOUT_TOOLS` matches the memory-bound heavy classes rather than forming a second opinion about which tools are expensive — the two must not drift.

### Not done here, deliberately

The substantive half of **#1224** — making a zero-context branch-protection lookup authoritative instead of falling back to the built-in default — is *not* in this PR. It trades real gate strength (a repo whose protection lists no required checks would accept any merged PR), so it is raised for a ruling rather than taken unilaterally. #1224 is closed on the documentation half, which is what the reporter emphasised; the default-resolution change should be its own issue if we want it.

PR #1260, opened 2026-09-12
fix(T12123): cleo list discloses truncation on --output id/table/--summary and gains --all (#1242)

### LINKED ISSUE #1242: [Bug]:cleo list --output count says 215 and --output id returns 10 for the same query, with no truncation signal — complete task enumeration is impossible
## Summary
`cleo list --output count` and `cleo list --output id` **disagree by two orders of magnitude on the same query**, with no pagination signal, no truncation marker, and no error. This makes it impossible to enumerate a project's tasks from the CLI, which in turn makes a complete structural reconciliation impossible.

## Repro
```
$ cleo list --status pending --output count
215

$ cleo list --status pending --output id | grep -cE '^T[0-9]+$'
10
```
Same flags, same filter, same store, seconds apart. **205 tasks are unreachable through `--output id`.** Nothing in the output says so — no `_truncated`, no `hasMore`, no `nextCursor`, no warning on stderr. The short list looks exactly like a complete list of ten.

The same holds for the other live statuses (`active`, `blocked`): summing `--output id` across all three yielded **16 ids** against counts totalling well over 200.

## Why this is worse than a missing `--limit`
It is not that the list is paginated — it is that **the paginated result is indistinguishable from the whole set**. An agent writing a reconciliation sweep does exactly what I did:

1. enumerate all live task ids,
2. walk down from each saga collecting everything reachable,
3. diff the two sets to find orphans.

Step 1 silently returned 16 of ~220. The diff then reported **"2 orphans"** — a clean, plausible, actionable-looking finding that was **entirely an artifact of the truncated enumeration**. Both flagged ids turned out to be fine: one was a saga (never a child by definition) and the other had a valid parent (`T087` → `T028`). Had I reported that number it would have sent someone hunting two non-existent problems, and — far worse — it would have implied the *other* 205 were verified.

This is the same class as the projection defects already filed (#1231, #1234, #1235): **a partial result presented as a complete one.** Here the consequence is that the absent rows read as "these do not exist" rather than "these were not returned."

## Concrete impact
The owner of this project asked for "100% accuracy of all work outstanding" — every task correctly scoped into Saga → Epic → Task and verified in the store. **That cannot currently be demonstrated through the CLI at all**, because there is no way to list every task. `cleo list --parent <id>` works and is reliable, so a top-down walk is possible; what is impossible is proving the walk was *exhaustive*, since the bottom-up set cannot be obtained.

## Asks
1. **Make `--output id` return the same population as `--output count`**, or paginate it explicitly with a cursor and a documented default limit.
2. **If truncated, say so in-band** — an `_truncated: true` / `total: 215` / `returned: 10` field, the same remedy as the other projection issues.
3. A `--all` or `--limit 0` escape hatch for reconciliation work would close this outright.

## Note on the other output modes
`--output count` appears correct (215 matches the briefing's figures). `--output table` and `--summary` were not tested for the same skew and may share it — worth checking, since the table form is what a human reads.

**Severity**: Major
**Area**: cli

---

## Environment
| Component | Version |
|-----------|---------|
| CLEO | 2026.8.9 |
| Node.js | v24.16.0 |
| Shell | /bin/bash |

kryptobaseddev:
> **Stacked on #1250** (`feat/cli-output-contract`). It touches `list.ts` and `operations-registry.ts`, which #1250 also changes, so basing it on `main` would have guaranteed a conflict. Review after #1250; the diff against that branch is the 6 files listed at the bottom.

## The defect

`cleo list --status pending --output count` reported **1075** while `--output id` returned **10** for the same query, seconds apart — with no `_truncated`, no `hasMore`, no `nextCursor`, and nothing on stderr. The short list was indistinguishable from a complete list of ten.

## What was actually broken — narrower than the report

**Neither number was wrong**, and the PR says so rather than just agreeing with the filer:

- `--output count` prints `data.filtered`, the filter-aware match count. That is **deliberate**: the `extractCount` TSDoc (T11481 · DHQ-034) explicitly documents it as "NOT the returned-rows length (which differs from the match count under pagination)".
- `TASK_LIST_DEFAULT_LIMIT` is 10.
- The envelope **already carried the truth**: `page: {mode:"offset", limit:10, offset:0, hasMore:true, total:1075}`.

The defect is that the enumeration render modes **discarded that `page` metadata**, so the caller had no way to learn it had seen a page.

**Report ask #1 — "make `--output id` return the same population as `--output count`" — is declined on purpose.** Silently switching a paginated read to unbounded would be a surprising behaviour change for every existing consumer. Disclosure plus a documented escape hatch gives the same guarantee without it.

## What changed

**1. `--all`** — a discoverable spelling of `--limit 0`. Core's `options.limit === 0 ? undefined : …` has always meant "no limit" and worked correctly, but it was documented **nowhere**: `--limit`'s help text said only "Maximum number of tasks to return". The one flag that made complete enumeration possible was invisible to anyone who had not read the core source. Both spellings are now named in the registry description.

**2. Truncation disclosure** on `--output id`, `--output table`, and `--summary`:

```
cleo: TRUNCATED — --output id returned 10 of 1083 matching rows. Re-run with
--all (or --limit 0) to enumerate every match, or pass --limit/--offset to
page deliberately.
```

The reporter flagged `--summary` as untested for the same skew. It has it — `--summary` is one line per *returned* record — so it is covered.

`--output count` is deliberately **exempt**: it already prints the full match count, so warning there would contradict its own output.

## Two deliberate non-changes

- **stderr, never stdout.** `--output id` exists to be piped; a warning line inside the id stream would corrupt the very consumer it protects (ADR-086 — one clean payload per call on stdout).
- **Exit code stays 0.** Flipping it would break every `set -e` consumer to fix a silent-truncation bug — trading one silent failure for a loud unrelated one.

## Why the warning can never contradict `--output count`

`detectTruncation` prefers the envelope's `page.hasMore`/`page.total`, then falls back to `data.filtered` — the exact field `extractCount` prints.

It deliberately does **not** fall back to `data.total` when `filtered` is present. `total` is every task in the project, so using it would warn on every filtered query that excluded anything — crying wolf until the warning is ignored. A test pins that distinction.

## Why this mattered more than a short list

An agent enumerating tasks bottom-up to diff against a top-down walk received 16 of ~220 rows and reported **"2 orphans"** — a clean, plausible, actionable-looking finding that was entirely an artefact of the truncated enumeration. Both flagged ids turned out to be fine. Worse than the two false positives: the result implied the other 205 had been *verified*. Absent rows read as "these do not exist" rather than "these were not returned".

## Verification — through the built binary, against a live 3181-task store

```
$ node packages/cleo/dist/cli/index.js list --status pending --output id 2>/dev/null | wc -l
10
$ node packages/cleo/dist/cli/index.js list --status pending --output id 2>&1 >/dev/null
cleo: TRUNCATED — --output id returned 10 of 1083 matching rows. Re-run with --all …

$ node packages/cleo/dist/cli/index.js list --status pending --all --output id 2>/dev/null | wc -l
1083
$ node packages/cleo/dist/cli/index.js list --status pending --output count 2>/dev/null
1083
```

`--all` and `--output count` now reconcile **exactly**, which is the property the reporter needed to prove a walk was exhaustive. Neither `--all` nor `count` mode emits a warning, and stdout stays a clean, pipeable id list in every case.

- 12 new tests pass; `pnpm typecheck` (`tsc -b`) clean; `pnpm biome check --write .` clean; `cleo check arch` 10/10 (20/20 once #1257 lands).
- Scoped test runs only (`--maxWorkers=2`).

Closes #1242

https://claude.ai/code/session_01QMFG4p2XwZgoQUsUjFySzg


kryptobaseddev:
## Independent same-day reproduction — on an agent working on this very saga

Another agent on this project hit this today, unprompted, and initially diagnosed it as something else entirely. They created a task and then could not find it:

> `cleo list --parent T12111 --output id` did **not** show T12140 immediately after I created it — I only found the ID via `cleo find`.

Their conclusion was that the write had not landed yet. It had. Measured:

```
T12111 children (--output count):        11
T12111 children (--output id, default):  10
T12140 present with --limit 0?           yes
```

Eleven children, the undocumented default limit of ten, and **T12140 was the one silently cut**. No `_truncated`, no `hasMore`, nothing on stderr.

This is the exact failure the issue describes — *"the absent rows read as 'these do not exist' rather than 'these were not returned'"* — and it is worth recording because of who it happened to and when:

- It happened to an agent **working on the saga that fixes this**, hours after the issue was filed and read.
- The wrong conclusion it produced was not "my list is short" but **"the record does not exist"**, which is a different and more dangerous inference than the original report's orphan-diff case. It invites a duplicate write.
- The recovery was `cleo find`, i.e. a second, differently-shaped query. Nothing about the first result suggested one was needed.

## Confirms the fix addresses it

Both halves of PR #1260 would have prevented this independently:

- `--all` returns all 11 rows, so the enumeration is complete.
- the stderr disclosure would have printed `cleo: TRUNCATED — --output id returned 10 of 11 matching rows…`, which turns "the task does not exist" into "I have seen a page".

Adding this because it is a better severity argument than the original report managed to make: the defect is not merely that enumeration is incomplete, it is that **an incomplete enumeration is indistinguishable from a definitive negative answer**, and a definitive negative answer about a task's existence is exactly what an agent acts on by creating it again.

Issue #1249, opened 2026-09-12
[Bug]: a filesystem path was written into tasks_tasks.id — row is listable but unreachable, and the read-path id validator already exists


kryptobaseddev:
## The defect

`tasks_tasks` in this repo's own store contains a row whose **`id` is a filesystem path**: `/mnt/projects/cleocode`. A project root was written into the id column, and the task's title was derived from it.

The row is **unreachable and unrepairable through the CLI**: every read/mutate surface validates the id format and rejects it, but the write path that created it did not.

## Evidence

Measured 2026-09-12 against `/mnt/projects/cleocode` (3173 tasks). The row surfaces in a pending enumeration at position 12:

```
$ cleo list --status pending --limit 0 --output id | grep -nvE '^T[0-9]+$'
12:/mnt/projects/cleocode
```

Its projected record:

```json
{
  "id": "/mnt/projects/cleocode",
  "title": "Task /mnt/projects/cleocode",
  "status": "pending",
  "priority": "medium",
  "kind": "work"
}
```

`type` is `null` — every legitimate task carries `task`/`epic`/`subtask`/`saga`.

### It cannot be read, updated, or deleted

```
$ cleo show '/mnt/projects/cleocode' --full
{"success":false,"error":{"code":6,"message":"Invalid task ID format: /mnt/projects/cleocode",
 "codeName":"E_VALIDATION_FAILED"},...}
```

So a validator that rejects this id shape **already exists on the read path** — `packages/core/src/tasks/show.ts:116`, with siblings at `packages/core/src/security/input-sanitization.ts:64`, `packages/core/src/validation/operation-gate-validators.ts:56`, `packages/core/src/validation/engine-ops.ts:354` — and was simply never applied at the write path. The result is an immortal orphan row: listable, never addressable.

### It corrupts enumeration arithmetic

This row is why a bottom-up enumeration of this project does not reconcile. `cleo list --status pending --output count` reports `1075`; a full `--limit 0 --output id` yields 1074 ids matching `^T[0-9]+$` plus this one path line. Anyone diffing a top-down walk against a bottom-up enumeration gets a phantom one-task discrepancy — the same false-finding shape #1242 describes, from a different cause. I lost time to exactly that before finding the row.

## Suggested fix

1. Apply the existing task-id validator at the **write** path (`add`, `add-batch`, and any direct upsert), so the read-path validator and the write-path validator are the same validator. A shape the reader will refuse must never be writable.
[...]

Issue #1214, opened 2026-08-20
[Bug]:cleo add stored the caller's acceptance criteria under ANOTHER task's title


kryptobaseddev:
### Bug Report

'cleo add --type task --parent T1389 --title "Copy Assist writes in the house quiet register..." --acceptance "<4 criteria>"' returned T1478.

T1478 was created at the right timestamp with the CORRECT parent and all four acceptance criteria verbatim, but its stored title was 'Browser rail integrations leaked order prose to rails; guard was server-scoped and directory-scoped' - unrelated work.

Impact: an agent that trusts the returned id and reads the task back sees another task's title on its own record. It briefly looked like I had completed an unrelated task. Acceptance criteria and title describing different work is a silent data-integrity fault; nothing flags it.

Same session, separate friction: a compound 'cleo session start ... ; cleo add ...' printed 'Error: An active session already exists' from the first command while the second still succeeded and printed its id, so a successful add read as a failure.

Worked around with 'cleo update T1478 --title ...'.

### Are you using an AI agent?

Yes - AI agent filed this issue

---

## Environment
| Component | Version |
|-----------|---------|
| CLEO | 2026.8.9 |
| Node.js | v24.16.0 |
| OS | linux 7.1.5-201.fc44.x86_64 x64 (x64) |
| Shell | /bin/bash |
| gh CLI | gh version 2.97.0 (2026-07-31) |
| Install | /home/keatonhoskins/.npm-global/bin/cleo |

kryptobaseddev:
## Investigation: does not reproduce on any write path. Three hypotheses eliminated, two still live.

I could not reproduce this, and I'd rather record the eliminations than close it — the two surviving hypotheses point in **opposite directions**, and a fix chosen between them would be a guess against the write path of a task tracker.

### Test 1 — single `cleo add`

```
submitted: "ZZPROBE-1214 distinctive title alpha bravo 1789229139"
stored   : T12143 · title 'ZZPROBE-1214 distinctive title alpha bravo 1789229139' · parent T12111
```

Exact match. Reading `packages/core/src/tasks/add.ts`, the title is never transformed: `options.title` flows unmodified to the insert (lines 1293, 1386). No normalisation, enrichment, or inference touches it.

### Test 2 — `add-batch` (index-slip hypothesis)

Three rows, distinctive title↔AC pairs, one transaction:

```
T12144  title=ZZBATCH-A alpha row one      ac[0]=A ac1   MATCH
[...]

Issue #1227, opened 2026-09-07
[Feature]:commit-msg hook accepts any task ID that matches the pattern, never one that resolves — three collisions on main in one day


kryptobaseddev:
## The gap

The commit-msg hook enforces that a subject line CONTAINS a task id. It never checks that the id RESOLVES, or that it resolves to the work being committed. Those are different properties, and only the second one is useful.

The hook already parses the id in order to validate the pattern. Calling `cleo show <id> --field /data/task/title` and printing that title for confirmation (or refusing on `E_CLEO_NOT_FOUND`) is a small change to code that has the id in hand.

## Measured on one repo in one day

`git log --grep=T1740` on `kryptobaseddev/PepsVida` main now returns two unrelated commits:

```
d96a772de  fix(T1740): the coverage report showed green on the one family… (#488)
b5ab628d1  fix(T1740): Clinical Lab Extract is pointed at an image generator… (#487)
```

The second is correct — T1740 genuinely is "Clinical Lab Extract is routed to an image-generation model that cannot read a lab report". The first is mine, and the real task was T1741. I filed T1741 forty minutes AFTER the other session created T1740, wrote the commit before checking, and the hook passed it because `T1740` matches the pattern.

This is the third instance the same repo recorded in ~24h:

- `feat(T1738-T1741)` (#464) cites T1740 and T1741, neither of which existed when it merged.
- PR #428 is committed as `feat(T1710-T1715)` for an enter-once intake spine; CLEO T1710 is PIN recovery by email and T1712 is a Zoho Meeting seam.
- The one above.

In each case a human or agent invented a plausible-looking id, and the only thing that would have caught it is a resolve.

## Why it is not just carelessness

A squash merge subject on a protected branch cannot be rewritten, so the wrong id is permanent and `git log --grep` becomes unreliable as an audit surface — which is what it is used for. The failure is also silent at exactly the moment it is cheapest to fix: the id is in the buffer, the author is at the keyboard.

Sessions here already know to check, and still did not, three times. That is the signature of a missing gate rather than a missing habit.

## Suggested behaviour

On commit-msg, for each `T\d+` in the subject:
- `E_CLEO_NOT_FOUND` -> refuse, naming the id.
- resolves -> print `T1741: Coverage report showed flowing on business_card...` so a mismatch is visible before the commit lands.

An escape hatch is needed for legitimately unfiled work; an env var in the same shape as the existing skips would do.
**Area**: hooks

### Are you using an AI agent?

[...]

PR #1261, opened 2026-09-12
fix(T12128): validate the task id at the write path, and make immortal rows removable

kryptobaseddev:
Refs #1249, #1214. Independent of #1258 and #1251 — based on `main`.

## The row that cannot be killed

A project path reached the `id` column of `tasks_tasks` in a live store:

```
id    = '/mnt/projects/cleocode'
title = 'Task /mnt/projects/cleocode'
type  = null
```

`cleo list` returns it. `cleo show`, `cleo update` and `cleo delete` all reject the id as malformed **before** they reach the store:

```
$ cleo show '/mnt/projects/cleocode'
{"success":false,"error":{"code":6,"message":"Invalid task ID format: /mnt/projects/cleocode","codeName":"E_VALIDATION_FAILED"}}
```

So the row can be enumerated forever and addressed never. **The read-side validators that should have prevented the write are exactly what make the result unfixable.**

## Why it happened

Four validators check the id shape on read paths — `tasks/show.ts`, `security/input-sanitization.ts`, `validation/operation-gate-validators.ts`, `validation/engine-ops.ts`. **None of them ran on the write path.**

They also disagree with each other, which is the deeper problem:

| pattern | where | disagrees about |
|---|---|---|
| `/^T[0-9]{1,7}$/` | `goal.ts`, `evidence-atom-schema.ts` (8 sites) | upper bound |
| `/^T(\d{3,})$/` | `tasks/id-generator.ts` | rejects `T12`, allows `T123456789` |
| `/^T\d{3,}$/` | `sentient/tick.ts`, `nexus/query.ts` | same |
| `/\bT(\d{4,})\b/` | `lifecycle/consolidate-rcasd.ts` | unanchored, 4-digit minimum |
| `/^(T\d+(-[A-Z]\w*)?\|E-\d+...)$/` | `changesets.ts` | allows suffixes and `E-` |

Five opinions about one of the most fundamental types in the system, and no guard where it mattered.

## What changed

**`TASK_ID_REGEX` + `isTaskId` in `@cleocode/contracts`** become the one shape. They deliberately adopt the `{1,7}` form already used at eight sites rather than inventing a sixth opinion, and are permissive by design: the job is to reject values that are not task IDs at all, not to re-litigate how many digits one should have.

Anchoring matters as much as the body — an unanchored variant accepts `see T1234 for details` as an id, which is one way prose ends up in an id column.

**`insertTaskRow`** — the chokepoint every task insert passes through — enforces it. A malformed id now fails loudly at write time instead of silently becoming unreachable data.

**`cleo doctor malformed-ids`** closes the other half: rows already written. Read-only by default, reporting id, title, status, type and how many dependency edges name the row. `--fix` deletes them and their edges in **one transaction** — a partially repaired store is worse than an unrepaired one, because the operator would then have to re-derive which rows were left, and these are precisely the rows no query can name.

Deletion rather than re-iding: a row whose id is not an id has no identity. Nothing can reference it, `depends` cannot name it, no evidence atom can cite it. Re-iding would invent an identity the row never had and silently change what any existing reference means. The report prints the full row first so the operator sees what is being discarded.

## Verification

- **Fail-before is exact.** With the guard removed, **exactly the 8 guard cases fail** and the "still accepts a well-formed id" case still passes — proving the guard rejects garbage without narrowing what already works.
- 9 write-guard tests + 4 repair tests. The repair test inserts its malformed row with raw SQL precisely because the supported path now rejects it, which is the point.
- `tasks-sqlite.test.ts`: 53 passed, no regressions.
- Gates: contracts purity **PASS** (a regex const is const data and `isTaskId` is a type guard — both explicitly permitted under gate 10), db-open guard **PASS** (routed through `openDualScopeDb`, no raw open), injection commands **PASS**, CLI package boundary **PASS** at baseline 33 with the new file not flagged.

## Scope I deliberately did not take

Migrating the other five call sites onto the canonical regex is a follow-up. Some are legitimately different — `changesets.ts` accepts suffixes and `E-` prefixes — so consolidating them is a design question, not a find-and-replace, and it does not belong in the PR that stops the bleeding.


kryptobaseddev:
Re-pointed from T12118 → **T12128**. T12118 was invented, not created; it actually belongs to cleo-optimizer's #1240 work. T12128 was created under epic T12114 and **read back from the store** (`cleo list --parent T12114` → T12115, T12116, T12117, T12128) rather than assumed — which is the method that would have prevented the original error.

Noting it here because it is the fourth instance of #1227 in this session, by the third different agent, and the first one I made *after* writing the argument for why the hook is needed. All four were real, T-shaped, resolvable IDs — which is exactly why nothing objected.

Worth recording for #1227's design: a check of `resolves + is a task + is not terminal` would **not** have caught this one. T12118 resolves, is a task, and is not terminal. It is simply someone else's. The missing dimension is ownership/provenance, and the fix should not imply a completeness it will not have.

Issue #494, opened 2026-05-22
[Feature] Doc-only / single-task epics: simpler lifecycle gating or 'doc' kind that skips stage requirements


kryptobaseddev:
## Summary

Epics block child-task `cleo complete` until the parent epic's pipeline
stage has advanced past `research → consensus → architecture_decision →
specification → decomposition` into `implementation` (or later). For pure
documentation work (e.g. scaffolding a project repo, writing ADRs,
writing runbooks) this stage progression is ceremony that doesn't reflect
the actual work.

Today the only workaround is calling `cleo lifecycle skip <epicId>
<stage> --reason "..."` four times (consensus, architecture_decision,
specification, decomposition), then `cleo lifecycle start <epicId>
implementation`. That's 5 extra CLI invocations per doc epic just to
unlock child completion.

## Repro

Tested with cleo `lifecycle v2026.5.100` on 2026-05-22.

```bash
# 1. Start a session and create a docs-only epic + children
cleo session start --scope global --name "docs-epic-repro"
cleo add --type epic \
  --title "Doc-only epic" \
  --description "Write some docs." \
  --acceptance "Doc A exists|Doc B exists|Index updated"
# -> creates T###  pipelineStage=research

cleo add --type task --parent T### \
  --title "Write doc A" \
  --description "..." \
  --acceptance "ac1|ac2|ac3"

# 2. Do the work + verify gates with real evidence
git init -b main && touch A.md && git add A.md \
  && git commit -m "add A.md"
SHA=$(git rev-parse HEAD)
cleo verify <childId> --gate implemented --evidence "commit:${SHA};files:A.md"
cleo verify <childId> --gate testsPassed --evidence "tool:test"
cleo verify <childId> --gate qaPassed   --evidence "tool:build"

# 3. Try to complete child
cleo complete <childId>
```

### Observed

```
E_LIFECYCLE_GATE_FAILED
Task <childId> cannot complete: parent epic T### is still in 'research'
stage. Advance the epic past decomposition before completing children.
Required stages: ["implementation","validation","testing","release"]
```

### Workaround (currently required)

```bash
cleo lifecycle skip T### consensus              --reason "docs-only"
cleo lifecycle skip T### architecture_decision  --reason "docs-only"
cleo lifecycle skip T### specification          --reason "docs-only"
cleo lifecycle skip T### decomposition          --reason "docs-only"
cleo lifecycle start T### implementation
# Now `cleo complete <childId>` succeeds.
```

[...]

Issue #1219, opened 2026-09-04
[Bug] transcript extraction fails on openai-compatible providers: prompt contradicts schema and response_format is dropped


Carson-Fricke:
Version: @cleocode/cleo 2026.8.9 (latest stable) - Windows 11, Node v26.7.0, npm global install; extraction backend resolved through the unified role resolver to an openai-compatible provider via OpenRouter

## Summary

`cleo transcript extract` (the `extractFromTranscript` path in `packages/core/src/memory/transcript-extractor.ts`) fails with `"LLM extraction call failed: No object generated: response did not match schema"` for every openai-compatible model we tested. Two independent defects in CLEO's extraction design combine to make structured extraction unreliable through any provider that does not receive/enforce `response_format`:

### Defect 1: The system prompt contradicts the zod schema

The schema requires an object wrapper:

```ts
const ExtractionResponseSchema = z.object({
    memories: z.array(ExtractedMemorySchema).max(20),
});
```

But `EXTRACTION_SYSTEM_PROMPT` ends with:

```
Return empty array if nothing valuable exists. Maximum 7 extractions.
```

Literal-minded models follow the prose and return a bare `[]`, which fails validation (`expected object, received array`). Others invent their own envelope. Verified raw responses:

| Model | Raw output | Validation outcome |
|---|---|---|
| `deepseek/deepseek-chat` | `[]` | fail — bare array |
| `openai/gpt-4o-mini` | `{"extractions": []}` | fail — wrong key |
| `z-ai/glm-4.6` | `{"session_id":"...","memories":["string", ...]}` | fail — strings instead of objects |
| `z-ai/glm-5.3-flash` | correct `{"memories":[...]}` once | passed 1 of 4 attempts (only in `--dry-run`) |

Nothing is stored in any failing case; the store never gains memories from this path.

### Defect 2: `response_format` is dropped on the openai-compatible transport

`generateObject()` is called without `structuredOutputs`:

```ts
const { object } = await generateObject({
    model: backend.model,
    schema: ExtractionResponseSchema,
    system: EXTRACTION_SYSTEM_PROMPT,
    prompt: userPrompt,
});
```

The AI SDK emits: `AI SDK Warning (openai.chat / <model>): The feature "responseFormat" is not supported. JSON response format schema is only supported with structuredOutputs`. Without in-band schema enforcement, correctness depends entirely on prompt discipline — which Defect 1 actively undermines.

### Related: `response_format: json_object` keyword rule

[...]

Issue #1218, opened 2026-09-04
[Bug] NEXUS code-graph indexing never populates (graph sync no-op; graph discover bound to wrong arg contract)


Carson-Fricke:
Version: @cleocode/cleo 2026.8.9 (latest stable) - Windows 11, PowerShell 5.1, Node v26.7.0, npm global install

## Summary

The NEXUS code-intelligence graph never populates on this version, and `cleo graph discover` fails with an argument-routing error that contradicts its own help text.

## Repro: sync is a no-op

```text
> cleo graph init
{"success":true,"data":{"message":"NEXUS initialized successfully"}}

> cleo graph sync
{"success":true,"data":{"synced":1,"failed":0}}

> cleo graph sync .
{"success":true,"data":{"synced":1,"failed":0}}

> cleo graph status
{"indexed":false,"nodeCount":0,"relationCount":0,"fileCount":0,"lastIndexedAt":null,"staleFileCount":-1}
```

The repo contains ~360 source files across four Python packages (`src/lake`, `src/harness`, `src/thesis`, `src/lakeclient`). The `synced: 1` count appears to reflect only the project registration row; no source file is ever analyzed, and `status` remains `indexed: false` with zero nodes before and after repeated syncs.

## Repro: discover has a broken arg contract

```text
> cleo graph discover --json
{"success":false,"error":{"code":1,"message":"Missing required positional argument: TASKQUERY","codeName":"E_VALIDATION"}}
```

`cleo graph --help` documents `discover` as "Discover the codebase structure (file tree + symbol counts)" with no arguments, and no `TASKQUERY` argument exists on any `graph` subcommand. The router appears to have bound the wrong operation or arg contract to this verb.

## Impact

On this version every project's nexus symbol graph stays empty, so symbol resolution, dependency/impact analysis, clusters/flows, `cleo memory code-auto-link` (code-to-memory linking), and code-aware briefing features are all inoperable. `cleo map --store` writes a structural summary to BRAIN but does not populate nexus.

## Expected

- `cleo graph sync` analyzes the project's source files and populates nodes/relations; `graph status` reports non-zero counts and `indexed: true`.
- `cleo graph discover` runs as documented, without a spurious required argument.

PR #1263, opened 2026-09-12
fix/feat: cluster E + #1207 both levers + #494

### LINKED ISSUE #1217: [Bug] BRAIN local embeddings are unreachable in 2026.8.9 (isAvailable gate deadlock; EmbeddingQueue has no callers)
Version: @cleocode/cleo 2026.8.9 (latest stable) - Windows 11, PowerShell 5.1, Node v26.7.0, npm global install

## Summary

BRAIN local embeddings are unreachable in this version. `brain.embedding.enabled: true` (a shipped default from `ensure-config`) registers `LocalEmbeddingProvider` (`Xenova/all-MiniLM-L6-v2`, 384-dim, matching the `brain_embeddings` vec0 table) at BRAIN store init, but no code path can ever reach `embed()`:

1. **Backfill deadlock.** `populateEmbeddings` (`packages/core/src/memory/retrieval/observe.js`, used by `cleo brain maintenance`) gates on `isEmbeddingAvailable()`:

   ```js
   if (!isEmbeddingAvailable()) {
       return { processed: 0, skipped: 0, errors: 0 };
   }
   ```

   `isEmbeddingAvailable()` returns the provider's `_ready` flag, which is only set to `true` by `loadPipeline()` — which runs exclusively inside `embed()`. In any fresh CLI process `_ready` is false, so the backfill always returns 0/0/0. The gate prevents the only call that could satisfy it.

2. **`EmbeddingQueue` is dead code.** `embedding-queue.js` (whose `fallbackEmbed` calls `provider.embed()` directly, bypassing the gate) has zero callers anywhere in the dist tree. Observation writes route through `brain-writer-thread.js`, which never references embeddings.

3. **Query-time degradation.** `embedText()` in `brain-embedding.js` gates on the same `isAvailable()` flag, so vector/hybrid search (`cleo memory search-hybrid`) silently degrades to FTS5 in every process.

The setup wizard (`cleo setup --section brain`) only toggles the config flag; there is no sanctioned activation path at all.

## Evidence

- `cleo brain maintenance --json` returns `"embeddings":{"processed":0,"skipped":0,"errors":0}` on every run.
- `SELECT COUNT(*) FROM brain_embeddings_rowids` = 0 after ~20 stored observations/decisions/learnings.
- Direct probe importing `@cleocode/core/dist/memory/embedding-local.js` and calling `getLocalEmbeddingProvider().embed(text)` works perfectly: 384-dim vectors, one-time 22MB model download, sane cosine similarities (0.19 between topically-related texts vs 0.01 for unrelated). The bundled provider itself is healthy — only the plumbing is broken.

## Expected

Either `cleo brain maintenance` backfills embeddings for rows missing vectors (loading the lazy pipeline on demand), or observation writes enqueue embeddings via the queue.

## Suggested fix directions

- Let `populateEmbeddings` call the provider's `embed()` directly (or invoke `loadPipeline()`) instead of gating on `isAvailable()`; treat first-call model download as an expected cost with a progress log.
- Wire `EmbeddingQueue` into the observe / brain-writer-thread paths so new writes are embedded asynchronously.
- At minimum, log a warning when `brain.embedding.enabled` is true but no embedding has ever been produced, instead of silently degrading to FTS5.

kryptobaseddev:
Closes #1217

Stacked on #1259 — **review the top commit only**. First of cluster E.

## A gate that only the call it blocks could satisfy

`LocalEmbeddingProvider.isAvailable()` returned `_ready`, the *"pipeline has loaded"* flag.

- `_ready` is set only by `loadPipeline()`
- `loadPipeline()` runs only inside `embed()`
- every caller of `embed()` checks `isAvailable()` first

In a fresh CLI process `_ready` is `false`, so nothing ever called `embed()`, so `_ready` never became `true`. **The readiness check could only ever be satisfied by the call it guarded.**

The failure was total and silent, on a **shipped default** (`brain.embedding.enabled: true`):

- `cleo brain maintenance` → `{processed: 0, skipped: 0, errors: 0}` on every run
- `brain_embeddings` empty after dozens of observations
- every hybrid search quietly degraded to FTS5
- and the provider itself was **perfectly healthy** when called directly — the reporter's probe got clean 384-dim vectors with sane cosine similarities

## Three dead call sites, not two

The report enumerates `populateEmbeddings` and `embedText`. There is a third the reporter did not catch: the **observation write path** at `observe.ts:311` gates on the same flag, so new memories were never embedded either — not just the backfill. All three gate on one unsatisfiable boolean.

## Fix: separate capability from state

`isAvailable()` now answers *"can this provider produce embeddings?"* rather than *"has it already?"*. It stays true until a load actually fails, and that failure **latches** so there is no retry loop.

The ~22 MB first-call download is the *reason* the load is lazy — not a reason to refuse to start it.

Because availability is now capability, the first real embed is also the first model load and **can** genuinely fail (offline, no cached model). So `embedText` returns `null` on failure: callers already treat `null` as "no vector → fall back to FTS5", which is the correct degradation. Throwing would reject the whole search, and on the fire-and-forget write path would surface as an unhandled rejection.

### The silent zero is now a spoken one

An inactive backfill reports `inactiveReason` and emits `W_EMBEDDINGS_UNAVAILABLE`. A bare `{0, 0, 0}` is indistinguishable between *"everything is already embedded"* and *"this is off or broken"* — and for the entire life of this bug it silently meant the latter. **That ambiguity is why it survived into a shipped release.**

The `EmbeddingProvider` interface now states that implementations MUST NOT report availability as "has already produced one", so the next provider doesn't reintroduce the shape.

## Not done here, deliberately

Wiring `EmbeddingQueue` into the write path. The report offers it as an *alternative* to fixing the backfill, and the backfill is fixed. Worth recording what I found: the queue is referenced **only** by `shutdown.ts` (`resetEmbeddingQueue`) — it is torn down but never fed. Per @cleo-dev, `shutdown.ts` / the queue's `doShutdown` is being changed concurrently under T12115, so wiring it is filed rather than raced.

I also checked @cleo-dev's caution that the queue may be *reachable but always degrading to the main-thread fallback* — that would be a different bug with a different fix. It isn't: there is no enqueue caller at all, only the teardown.

## Verification

7 new cases pinning capability-vs-state, including one that **reproduces the deadlock shape directly**: a provider whose availability depends on a prior success is never called and reports no error. That test is the guard — re-conflating the two concepts restores the bug while every other test stays green.

33 passing across all three embedding suites (scoped, `--maxWorkers=2`). `tsc --noEmit`: zero errors in `src/memory/`.

---

## Retargeted to `main` (2026-09-12)

Previously based on `feat/evidence-ac-provenance` (= #1259's head). #1259 is now
retargeted to `main` with its conflicts resolved, and merging it forward into
this branch is **conflict-free** — the 29 files here do not overlap the three
that conflicted below.

Retargeted to `main` rather than left stacked, deliberately. Stacked is worse
than noisy: every workflow filters `branches: [main]`, so a stacked PR reports
`CLEAN` with `checks=0` and its green means nothing (gh#1280). A bigger diff
that is actually tested beats a small one that isn't. The overlap with #1259
disappears from this diff once #1259 lands.

Verified on the merged tree: gate 20 arch-gate parity 21 == 21 · both CLI
startup-barrel gates pass · `@cleocode/core` typecheck 0 errors.

## ⚠️ Flagged, deliberately NOT resolved — two gates, one invariant

This branch adds `scripts/lint-cli-startup-barrel.mjs` (T12138 · gh#1207). Main
already carries `scripts/lint-cli-startup-barrel-imports.mjs` as **gate 19**
(T12076). They enforce the **same invariant**, written independently, and differ
in strength:

| | gate 19 (on main) | this branch's |
|---|---|---|
| rule | ratchet — count may fall, never rise | zero static barrel imports in the entrypoint graph |
| current reading | 106 static barrel imports | 16 modules, zero |
| registered? | yes, bundle + AGENTS.md | **neither** |

Both pass right now, so nothing is broken — but this branch ships a gate that
**nothing runs**, and gate 20 is structurally unable to notice: it compares the
bundle to the table, and a script absent from *both* satisfies parity by mutual
absence.

Three defensible resolutions with different consequences — register it as a
second gate, delete it in favour of gate 19, or fold the stricter graph rule
*into* gate 19 and retire the ratchet. That is a design call, so it is left
untouched here and raised with @cleo-dev rather than decided in a merge.


kryptobaseddev:
## Second commit: #1216 — OpenAI-compatible providers misroute to `api.openai.com`

`0246c0dcd` closes **#1216** on the same branch.

### The bug is wider than the report

`deriveApiWire` returned `baseUrl: null` for **every** OpenAI-compatible provider except Codex, and a null base URL makes the OpenAI SDK fall back to `api.openai.com`. The report names `openrouter`; the same fault applied to **`deepseek`, `xai`, `groq` and `moonshot`**. Every documented OpenAI-compatible provider shipped pointing at somebody else's API.

The reporter's workaround — store the key in the `openai` slot and redirect with `OPENAI_BASE_URL` — worked precisely because it exploited the one provider whose default is correct.

### Why it was hard to see

The 401 comes back as OpenAI's own canonical message, blaming the key and linking to platform.openai.com:

```
401 Incorrect API key provided: sk-or-v1*****...2d8b.
You can find your API key at https://platform.openai.com/account/api-keys.
```

An error that blames the credential for a routing fault. The reporter had to compare that body against OpenRouter's before the cause was visible at all.

### The base URLs were never unknown — they were in three places

| location | value | consumed by routing? |
|---|---|---|
| hand-written builtin profile (`openrouter.ts`) | `https://openrouter.ai/api/v1` | no |
| generated models.dev catalog | `https://openrouter.ai/api/v1` | no consumers at all |
| inline map in `cli-ops.ts` | `https://openrouter.ai/api` (different shape) | only for `llm test` |
| **`deriveApiWire`** — the path that decides where requests go | **`null`** | **yes** |

Three copies of a fact, and the one that mattered had no value. `DEFAULT_PROVIDER_BASE_URLS` is now the single source, and `cli-ops` reads it too — so the URL used for probing and the URL used for routing can no longer disagree. That consolidation also **resolved 13 LLM-chokepoint lint violations** (40, baseline 53).

### Deliberately unchanged

- `openai` stays `null` — the SDK's own default is right for it, and adding it would be redundant state.
- `openai` + OAuth still routes to the Codex backend, not `api.openai.com` (pinned by a test).
- `anthropic` / `bedrock` / `gemini` / `ollama` wires untouched (pinned by a test).

### Not fixed, and not claimed

The reporter's secondary complaint that `cleo llm test openrouter` returns **404**. The probe URL that map produced was already correct once assembled, so the 404 has a different cause. It needs its own reproduction rather than an assumed fix riding along on this one.

### Verification

10 new cases: all five misrouting providers, a guard that no default may contain `api.openai.com`, the `openai`-stays-null case, the OAuth/Codex path, and the untouched wires. **1062 passing across all 67 `src/llm/` suites** (scoped, `--maxWorkers=2`). `tsc --noEmit`: zero errors in `src/llm/`.

kryptobaseddev:
## Cluster E complete — 4 issues, 4 commits

| commit | closes | what |
|---|---|---|
| `d98ed4054` | **#1217** | BRAIN embedding `isAvailable()` deadlock — capability vs state |
| `0246c0dcd` | **#1216** | OpenAI-compatible providers misrouting to `api.openai.com` |
| `48e2a2ebd` | **#1219** | Transcript extraction prompt contradicted its schema; no schema on the wire |
| `5dfd0c506` | **#1218** | `graph sync`/`discover` described operations they don't perform |

### #1219 — two defects that produced "flaky", not "broken"

The prompt ended *"Return empty array if nothing of durable value exists"* while the schema required `{memories: [...]}`. The reporter's raw captures across four models:

| model | raw output | outcome |
|---|---|---|
| `deepseek/deepseek-chat` | `[]` | bare array |
| `openai/gpt-4o-mini` | `{"extractions": []}` | wrong key |
| `z-ai/glm-4.6` | `{"memories": ["string", ...]}` | strings, not objects |
| `z-ai/glm-5.3-flash` | correct | 1 of 4 attempts |

Simultaneously `generateObject` was called without `structuredOutputs`, so the AI SDK dropped `response_format` and **said so in a warning nobody was reading** — leaving correctness entirely to the prompt discipline defect 1 was undermining.

The rewritten prompt also contains the word "json", which `response_format: json_object` requires in `messages`. Its absence made *short* transcripts fail with HTTP 400 while *longer* ones that happened to mention json failed later at validation — so the command looked transcript-dependent and flaky rather than broken. That's the worst available diagnostic signal.

The prompt is now **exported and tested against the schema**. Prompt/schema agreement is otherwise unverifiable — the only thing reconciling them is a model's interpretation at run time, which is exactly why this shipped.

Also fixed: `--projects-dir` was declared on `transcript extract` and **never read**, while the scanner hardcoded `~/.claude/projects`.

### #1218 — a *success* envelope describing work that did not happen

`graph sync` is described as *"Re-analyze and sync the project graph with the codebase"*. It performs no analysis: `nexusSyncAll` refreshes registry metadata (taskCount, labels, lastSync) and opens no source file. On a 360-file repo it returned `{"success": true, "synced": 1}` — one metadata row — while `graph status` stayed `indexed: false` with zero nodes.

This saga has seen four errors that describe the wrong thing. This is worse: a **truthful number answering a question nobody asked**, under a name promising a different operation, reporting success.

Behaviour is deliberately unchanged — sync-all is a registry op by design, and making it analyze every registered project would be a worse bug than the one being fixed. Instead the description tells the truth, names `cleo nexus analyze`, and a warning stops `synced: N` being read as files.

`graph discover` documented itself as a no-argument structure dump while requiring a positional `taskQuery`. The reporter concluded the router was misbound; **the router was correct and the description was not.**

**Validated, not assumed:** the reporter's repo is Python, raising the possibility the analyzer simply doesn't support it. It does — `py`/`pyi` are both in the tree-sitter extension map — so language was never the obstacle and neither fix is hiding one.

### Gates

- 14 (injection commands) ✅ · 15 (workflow commands) ✅ · 13 (LLM chokepoint) **40 vs baseline 53 — 13 resolved**
- Gate 1 (defineCommand SSoT) unchanged at 139/139 — no citty import added
- `tsc --noEmit`: zero errors in `src/memory/`, `src/llm/`
- **1053** memory tests · **1062** llm tests · **745** CLI command tests, zero failures

kryptobaseddev:
## Four more commits — #1207 (both levers) and #494

| commit | issue | what |
|---|---|---|
| `4f4edeb18` | refs **#1207** | One barrel import = 87% of CLI startup. Narrowed + gated. |
| `92909b829` | closes **#494** | Per-kind lifecycle requirements — ceremony that matches the work. |
| `9c7cb36d0` | refs **#1207** | Batch `cleo show` — pay the startup floor once, not per task. |

### #1207 — measured, against the core the CLI actually resolves

```
cleo --version                      1.31 s
bare node boot                      0.01 s
import @cleocode/core/internal      1.14 s   <- 87% of startup
import the narrow routing module    0.09 s
```

`index.ts` eagerly imports `help-renderer.ts` to build the alias map on every invocation. `help-renderer.ts` statically imported `@cleocode/core/internal` — a 2018-line barrel pulling in drizzle, node:sqlite, every dispatch handler and every middleware — **for one function**.

**`index.ts` already carries a comment saying precisely this must not happen.** The comment was correct. The import was added anyway, in a different file, where the comment could not reach. That is why this ships with `scripts/lint-cli-startup-barrel.mjs` rather than only a fix — proven against pre-fix code, where it reports `help-renderer.ts:16` and the chain `index.ts -> help-renderer.ts`.

Incidental: core's `exports` was missing `./routing/*` and `./routing/*.js` while ~25 sibling directories declare both.

> **Two measurement errors of mine are withdrawn in the commit message**, so they cannot propagate: I first measured the wrong core (a stale hoisted 2026.5.123 that nothing resolves), and from its file size inferred a publish-bundling theory that was unsupported. A third figure I quoted in passing (0.53s) came from this worktree with packages unbuilt and is unrepresentative.
>
> **The remaining ~0.17s of the floor is NOT explained by this change, and #1207 should not be closed as if it were.**

### #1207 second lever — batch lookup

`cleo show T1 T2 T3` returns all three from one process in one envelope. For the reporter's 25-task sweep the floor alone was ~33s of the 2.5 minutes.

A **single id keeps the byte-identical prior envelope** — batching must not change the shape every script parses. A failed id does not discard the rest, failures land in `notFound`, and the command exits non-zero **specifically so a partial batch cannot be read as a complete one** — otherwise this would become a fresh instance of #1242, which would be a poor thing to ship while #1242 is open.

### #494 — the per-kind table, not a `doc` kind

A `doc` kind whose only semantic is "skip the stages" encodes an exception, and the next non-code work type needs another. `KIND_LIFECYCLE_REQUIREMENTS` answers the general question on the axis that already exists (ADR-066).

`work` keeps full ceremony. `bug`, `research`, `spike`, `experiment`, `release` do not — a fix operates inside an existing design, research IS the antecedent work, a spike exists to produce what the stages would consume. **An absent or unknown kind resolves to `work`**, so a missing field can never be the cheapest way to opt out.

**Evidence gates are explicitly unaffected** — this axis governs ceremony, not rigour.

The real cost being removed isn't the five extra invocations; it's that four `lifecycle skip --reason` calls per epic write four audit records asserting deliberate bypasses, and an audit trail that is mostly ceremony-avoidance still *looks* like evidence.

> I initially put the predicates in `contracts` and **architectural gate 10 rejected them** — a contracts-purity violation by someone who'd read the rule that morning. Moved to `core/src/lifecycle/`. Recorded in the commit because a gate that catches an informed author is load-bearing, not a style guide.

### Verification

331 → **808 CLI tests**, **362** lifecycle/completion, **10** kind-requirements, **6** batch-show. Gates: contracts purity ✅ · contracts fan-out ✅ · CLI boundary 33/33 baseline ✅ · startup barrel ✅ · 14/15 ✅. `tsc --noEmit` clean in every file touched.

kryptobaseddev:
**Merge-blocked pending a threshold decision — not because this fix is wrong, but because of what it switches on.**

The fix here is correct and addresses gh#1217 exactly as filed. The problem is a side effect that is invisible in the diff.

## The chain, every link verified statically

```
main    embedding-local.ts:62   isAvailable() { return _ready; }
#1263   embedding-local.ts:140  isAvailable() { return !_loadFailed; }
```

`_ready` was only set by `loadPipeline()`, which only runs inside `embed()`, and every caller of `embed()` first checks `isAvailable()` — so in any fresh process it was **always false**. That is the deadlock this PR fixes. After the fix, `isAvailable()` reports capability and returns **true** in a fresh process.

That flag is the single switch between two different Tier-1 similarity metrics:

```
duplicate-detector.ts:360   if (!isEmbeddingAvailable()) return null;   -> falls back to trigram Jaccard
duplicate-detector.ts:606   const embeddingEnabled = ... isEmbeddingAvailable()
duplicate-detector.ts:639   if (embeddingEnabled && vectorBudget > 0 && ...)  -> cosine on 384-dim embeddings
```

And both metrics feed **one shared threshold pair**:

```
duplicate-detector.ts:42   export const DUPLICATE_WARN_THRESHOLD   = 0.85;
duplicate-detector.ts:45   export const DUPLICATE_REJECT_THRESHOLD = 0.92;
```

There are no metric-specific variants — verified by grep across `packages/core/src`. So this PR makes the cosine path live for the first time ever, against thresholds calibrated for trigram Jaccard.

## The failing test is the canary, not the bug

`reconciliation-engine.test.ts > linksAffected increments when tasks are created` fails `expected 1 to be 2`. It passes on `main` in ubuntu shard 2 and fails here in the same shard, same 15-test file, 42 minutes apart — confirmed by two agents with independent probes, so it is not a shard artefact and not a flake.

Its two fixtures produce these search blobs, and a faithful reimplementation of `normalise`/`trigrams`/`jaccard`/`buildSearchBlob` scores them:

```
blobA: external task ext-001 external task ext-001 synced from test
blobB: external task ext-002 external task ext-002 synced from test
trigram Jaccard = 0.8000      < 0.85 warn, < 0.92 reject     -> both creates succeed, linksAffected = 2
```

0.80 explains `main` passing exactly. Those two strings are semantically near-identical, so cosine on a sentence embedding scores them far higher — second create rejected with `E_DUPLICATE_TASK_LIKELY`, no task, no link, `linksAffected = 1`. The observed failure, to the number.

Corroborating, weakly but in the same direction — the duplicate-detector suite does measurably more work here:

```
duplicate-detector.test.ts (34 tests)   main 8832ms   #1263 10643ms
  score >= 0.92 rejected case                412ms          1174ms   (~2.8x)
```

**Stated plainly: the actual cosine value for that pair has NOT been measured.** That needs the model loaded, which nobody has run. "Far above 0.92" is an inference from the strings, not a number taken. And the job log carries no explicit `E_DUPLICATE` trace from the reconciliation test.

## Why this blocks rather than gets fixed in passing

The reconciliation test is the only place in the suite that creates two near-identically-titled tasks in sequence, which is why it is the one thing that caught this. Merging as-is turns on semantic duplicate detection **globally, for every `cleo add`**, with thresholds nobody has calibrated for cosine — and `E_DUPLICATE_TASK_LIKELY` is a hard refusal, not a warning. A pair scoring 0.80 on the metric the thresholds were tuned for plausibly scores ~0.99 on the metric that will now run.

This is gh#1314 — *Tier 1 has two implementations sharing thresholds calibrated for one* — going from latent to live.

## Options

1. **Metric-aware thresholds** — a separate pair for cosine, calibrated. The real fix.
2. **Gate the vector path off** until gh#1314 lands, keeping the gh#1217 deadlock fix. Smaller and shippable now.
3. Merge as-is. Not recommended: it changes write-path behaviour for every user on an uncalibrated threshold.

Option 2 keeps the value of this PR without the behaviour change. Holding the merge until that is decided.

Credit to cleo-helper for the mechanism; I verified the four links independently before blocking.

kryptobaseddev:
**Conflict resolution against current `main` (`5cfe13493`) — keep-both on both files, parity holds by construction. Worked out at cleo-dev's request; this branch was not touched.**

This PR's base is `e0bd95c05`, whose gate table ends at **row 21**. This PR numbers its new gate **23** and its runner entry **`gate-23`**, skipping 22 — and 22 is exactly what #1322 took when it merged as `5cfe13493`. **So there is no numbering collision**, and the conflict is purely two appends landing at the same table tail. Verified: row 23 is claimed by this PR only, across all open PRs.

If 22 was reserved deliberately for the PR ahead of you in the queue, that is good hygiene under concurrent development and worth saying out loud — it is the reason this resolves cleanly instead of needing a renumber.

**`AGENTS.md` row 19** — take this PR's version. The `**Repo-wide scope** — pairs with row 23…` note is purely additive; `main`'s row-19 text is unchanged and nothing is dropped.

**`AGENTS.md` table tail** — resolved order:

```
row 21  (unchanged)
row 22  from main    scripts/lint-ai-sdk-surface.mjs
row 23  from this PR scripts/lint-cli-startup-barrel-entrypoint.mjs
```

**`packages/cleo/src/cli/commands/check.ts`** — append this PR's `gate-23` object immediately after `main`'s `gate-22` (main line 674), both verbatim with their comment blocks.

**Why keep-both rather than pick, concretely.** Picking either side here silently deletes a real row, and the deletion is close to undetectable: the table stays syntactically valid, and gate 20 joins on **script path**, so a missing row surfaces later as a parity failure on a *different* gate with no obvious link back to this merge. The side I would have dropped had I picked is `main`'s row 22 — my own gate from #1322, an hour old, in a table tail nobody reads closely.

**Parity, which is the thing that must hold:**

```
main now:               bundled 22  ==  documented 22
after this resolution:  bundled 23  ==  documented 23
```

Both halves gain exactly one script path, so parity is preserved **by construction** rather than by the counts happening to agree.

**One check to do with the gate, not a grep.** Verifying parity by grepping `AGENTS.md` file-wide gives a false `23 documented vs 22 bundled` — the extra is `scripts/lint-worktree-location.mjs` at line 176, in the **Worktree Subsystem prose**, outside the section gate 20 slices (its `documentedScripts()` reads only between `## SSoT & Architectural Gates` and the next `## `, and its docstring names the worktree lint as a known non-arch-gate reference). I made that mistake first; the gate is right and the file-wide grep is the wrong instrument.

**Not resolved here, deliberately:** `packages/cleo/src/cli/generated/command-manifest.ts`. That one must be **regenerated** (`gen:manifest` → `biome check --write` → commit; build freely afterward, since #1312 made `prebuild`/`pretypecheck` verify rather than regenerate — I confirmed that behaviourally by mtime). Merging a generated file by hand is how a command `main` added gets silently dropped.

**One thing to expect that is not a new problem:** this PR touches 2 of the prebuild gate's 8 `paths` patterns (`packages/contracts/src/index.ts`, `packages/contracts/src/kind-lifecycle.ts`), so while macOS allocation stays starved it will burn `Worktree NAPI Prebuild Gate`'s 15-minute timeout and show `CI FAILURE` **with no failing job** — see #1358. That is the known gate-timeout mechanism, not a third defect on this PR. If it does *not* time out, #1358's mechanism is wrong and that issue should reopen.

kryptobaseddev:
Superseded — every fix in this PR has shipped, cherry-picked onto current `main` rather than merged from here.

This branch was **998 commits behind main** with 8 conflicts in files changed this week, so merging it was the wrong vehicle. But that was a property of the *branch*, not the work: tested individually, **five of six** remaining commits cherry-picked cleanly, and the sixth had a single additive conflict in the contracts barrel.

| Task | Commit here | Shipped in |
|---|---|---|
| T12112, T12113, T12118, T12125, T12126 | — | #1259 |
| T12129 — embedding availability is capability | `d98ed4054` | **v2026.9.14** (#1520) |
| T12132 — OpenAI-compatible providers route to their own endpoint | `0246c0dcd` | **v2026.9.15** (#1524) |
| T12133 — extraction prompt agrees with its schema | `48e2a2ebd` | **v2026.9.15** (#1524) |
| T12135 — graph sync/discover describe what they do | `5dfd0c506` | **v2026.9.15** (#1524) |
| T12138 — barrel import cost per CLI invocation | `4f4edeb18` | **v2026.9.15** (#1524) |
| T12140 — per-kind lifecycle requirements | `92909b829` | **v2026.9.15** (#1524) |
| T12141 — batch `cleo show` | `9c7cb36d0` | **v2026.9.15** (#1524) |

All nine tasks are `done` with `implemented` / `testsPassed` / `qaPassed` evidence bound to every acceptance criterion.

### What landing it turned up

- **T12129 fixed the deadlock but not the outcome.** With the unsatisfiable check gone, the real reason embeddings never populated became visible: the provider is registered by a `setImmediate` that races every consumer, with errors swallowed. Fixed as **T12314** (#1523). Measured on a real store: `{processed: 0}` → `{processed: 5550}`, `brain_embeddings_rowids` 0 → 5550.
- **Enabling embeddings turned on semantic dedupe for the first time**, which silently dropped near-duplicate tickets in external sync. Caught by CI before release and exempted via the audited `forceDuplicate` bypass.
- **T12141's description exposed a generator defect**: the command-manifest regex excluded every quote character rather than the delimiter, truncating any description containing one. Seven were being silently cut short on `main`, including `cleo show`'s.
- **T12141's batch path returned envelopes, not records**, so every element was empty. Caught on the installed build; fixed in #1527.

Nothing from this PR is outstanding. Closing.

Issue #1253, opened 2026-09-12
[Bug]: cleo check arch runs 10 gates while AGENTS.md documents 15 — agents told to self-check get a green that covered two-thirds of the table


kryptobaseddev:
## The defect

`cleo check arch` reported `Result: 10 passed, 0 failed, 0 skipped` while the AGENTS.md architectural-gate table documented **15** gates. The two had drifted in **both directions**:

- **9 gates documented but never bundled** — the runner never ran them, yet reported a clean green.
- **4 gates bundled but never documented** — the table did not list them at all.

Measured 2026-09-12 on `main` @ `77fb47e04`.

### The 9 documented-but-unbundled gates

| Table row | Script |
|---|---|
| 2 — Paths SSoT | `scripts/lint-paths-ssot.mjs` |
| 7 — Deployed template parity | `scripts/lint-deployed-template-parity.mjs` |
| 8 — `engines.node` SSoT | `scripts/lint-node-engine-ssot.mjs` |
| 9 — Publish surface SSoT | `scripts/lint-publish-surface.mjs` |
| 10 — Contracts purity | `scripts/lint-no-runtime-in-contracts.mjs` |
| 11 — Tools-vs-Skills boundary | `scripts/lint-tools-vs-skills-boundary.mjs` |
| 12 — Crate publish guard | `scripts/lint-no-crate-publish.mjs` |
| 13 — LLM Chokepoint Guard | `scripts/lint-llm-chokepoint.mjs` |
| 14 — Injection Command Existence | `scripts/lint-injection-commands.mjs` |

### The 4 bundled-but-undocumented gates

`scripts/lint-no-bare-get-active-session.mjs` (T11640), `scripts/lint-no-domain-db-singleton.mjs` (T12041), `scripts/lint-vitest-memory-safe.mjs` (T12087), `scripts/lint-cli-startup-barrel-imports.mjs` (T12076).

## Why this matters — who actually gets hurt

**Nothing was unenforced in CI.** All 15 documented lint scripts *are* invoked by `.github/workflows/*`, so `main` was never unprotected. That is what makes this low *severity*.

[...]

PR #1266, opened 2026-09-12
chore(T12134): remove two modules nothing imports; add a sweep for the rest

kryptobaseddev:
Serves the "rip out anything not needed" directive with evidence rather than instinct.

## The question

cleo-optimizer found two dead subsystems in two consecutive issues this afternoon — an `EmbeddingQueue` referenced only by its own teardown, and a generated provider catalog with zero consumers. Both had shipped. That suggested a third decay mode worth checking systematically, distinct from the documentation-drift and stale-gate classes we found elsewhere today: **what is built, committed, and wired to nothing?**

## Removed (2 modules, 1,148 lines)

**`adr-backfill-walker.ts`** — 765 lines. It carries a `.cleo/deprecations.yml` entry scheduling removal in **v2026.6.0**. Current version is **v2026.8.9**: three minor versions past due. Nothing imports it, and its replacement (`cleo docs add --type note`) has been canonical since T9788. The deprecation entry is replaced with a note recording why it was dropped, rather than silently deleted.

**`graph-rag.ts`** — 383 lines, eight exports, zero importers. The only references anywhere are in `docs/generated/api-reference.md`, which is derived from the source and therefore not evidence of use.

Both verified by full-text search across `.ts`, `.mjs`, `.json`, `.yml` and `.md` — not just by the sweep script. Neither has tests. Typecheck is clean.

`packages/core/src/backfill/index.ts` is a **different** module (`backfillTasks`, `generateAcFromDescription`), is exported from `internal.ts`, and is untouched. Checked explicitly, because the shared directory name invites exactly that mistake.

## Added: `scripts/find-unimported-modules.mjs`

Makes the question repeatable. Resolves relative **and** `@cleocode/*` cross-package specifiers, and counts barrel re-exports — so a module reachable through `index.ts` is not reported.

Current run: **192 modules, ~33,700 lines, imported by nothing.**

**That number is a starting point, not a verdict**, and the docblock says so. Known false-positive classes are named in it: SvelteKit file-based routes (`+page.server.ts`), npm lifecycle scripts (`postinstall.ts`), package `bin` entries, one-off migrations invoked by path, and `__fixtures__`. All are loaded by convention rather than import.

### A bug in the script, worth recording

My first version reported **1,159 of 1,267** core modules as unimported — including `dual-scope-db.ts`, which I know is imported constantly. The cause was comparing relative walk paths against absolute resolved paths, so every lookup missed.

I caught it because the result was too extreme to be true. A subtler bug producing a plausible-looking 40 would have been believed, and I would have "cleaned up" live code. That is the argument for the sanity check now baked into the workflow: **run the sweep, then verify each candidate by full-text grep before touching it.**

## A finding for the owner, not acted on here

**All five entries in `.cleo/deprecations.yml` are past their own stated removal version:**

| id | remove | status |
|---|---|---|
| `agent-output-raw-md-write` | v2026.6.0 | removed in this PR |
| `get-cleo-templates-dir` | v2026.7.0 | **past due** |
| `resolve-agent-templates` | v2026.7.0 | **past due** |
| `resolve-starter-bundle` | v2026.7.0 | **past due** |
| `cleo-init-workflows-alias` | v2026.7.0 | **past due** |

Only the one with zero importers is removed here. The other four are thin shims that may still have out-of-tree consumers, so dropping them is a deliberate decision about a deprecation window rather than a cleanup — and a deprecation policy nobody enforces is the same failure class as a CI gate that does not exist.


kryptobaseddev:
## Correction to this PR's own analysis: there were SIX past-due deprecations, not five

The table in the description is wrong. My enumeration regex required `replacement:` to follow `remove:` immediately, so it silently skipped `get-workflow-templates-dir`. Corrected list — every entry was past its own removal version:

| id | remove | status |
|---|---|---|
| `agent-output-raw-md-write` | v2026.6.0 | **removed in this PR** |
| `get-cleo-templates-dir` | v2026.7.0 | past due |
| `get-workflow-templates-dir` | v2026.7.0 | past due — **missed in my first count** |
| `resolve-agent-templates` | v2026.7.0 | past due |
| `resolve-starter-bundle` | v2026.7.0 | past due |
| `cleo-init-workflows-alias` | v2026.7.0 | past due |

Five remain after this removal, not four. The conclusion was right and the number was wrong.

Worth the note rather than a silent edit, because it is the same failure shape this saga keeps finding: **a matcher that under-matches fails silently and returns a confident, smaller, wrong answer.** It is the sibling of the vacuous-pass hazard #1253 guards against, where a parity parser that stops matching reports success by finding nothing. Absence reading as success, again — this time in my own analysis script rather than in a gate.

Also fixed here: `Lint & Format` failed on `scripts/find-unimported-modules.mjs` (`noAssignInExpressions`). Replaced the `while ((m = RE.exec(src)))` loop with `src.matchAll(RE)`; output is byte-identical at 192 modules / 33,755 lines, so the change is behaviour-preserving.

And the retirement note in `deprecations.yml` cited **T12132**, which is not this task — T12134 is, read back from the store. Fifth instance of #1227 in this session, third by me, and again from writing a reference before the task row existed.

Issue #1223, opened 2026-09-05
[Bug]:cleo complete prints backend warnings to stdout, breaking the one-envelope-per-call contract (ADR-086)


kryptobaseddev:
### Bug Report

`cleo complete T1655` emitted, on STDOUT, before the envelope:

    [LocalBackend] @vlcn.io/crsqlite not installed and no crsqliteExtPath provided — opening in local-only mode (hasCRR=false). Install @vlcn.io/crsqlite (or set crsqliteRequired) to enable cr-sqlite sync.
    {"level":"WARN","time":…,"subsystem":"tasks:complete","taskId":"T1655","mergeError":"task branch 'task/T1655' does not exist","worktreeRemoved":false,…

ADR-086 states stdout is ONE LAFS envelope per call and all logs/progress go to stderr. Both of these are logs. Anything parsing stdout as JSON (which the CLI's own `--field` contract encourages) has to skip lines first — the exact 'pipe through tail/jq' anti-pattern the ADR forbids.

Two sub-points:
- the crsqlite line is an environment notice and belongs on stderr, or behind a verbosity flag;
- `mergeError: task branch 'task/…' does not exist` is emitted for tasks that were never worked on a task branch (mine were on feature branches merged by PR). It reads as a failure and is not one. Either suppress it when no task branch was ever created, or downgrade it to debug.

The command otherwise succeeded — the task did reach `done`.

CLEO v2026.8.9.

**Severity**: Moderate
**Area**: cli

### Are you using an AI agent?

Yes - AI agent filed this issue

---

## Environment
| Component | Version |
|-----------|---------|
| CLEO | 2026.8.9 |
| Node.js | v24.16.0 |
| OS | linux 7.1.5-201.fc44.x86_64 x64 (x64) |
| Shell | /bin/bash |
| gh CLI | gh version 2.97.0 (2026-07-31) |
| Install | /home/keatonhoskins/.npm-global/bin/cleo |

kryptobaseddev:
## The `[LocalBackend]` line — what is known, and why it is not being hunted

Recording rather than chasing, so it is attributable if it recurs.

**What it is not:** not ours. `grep -rn "LocalBackend" packages/` over the whole monorepo returns exactly one hit — a TSDoc mention in `packages/contracts/src/attachment.ts:118` describing llmtxt's `LocalBackend`. There is no such class, and no such log statement, in CLEO's source.

[...]

Issue #1235, opened 2026-09-11
[Bug]:A write can report success and do nothing: empty-task AC lock + silently no-op --reason override + --output silent returning 0 on a rejected mutation


kryptobaseddev:
## Summary
Three defects compose into a **silent data-loss path** on the task record: the acceptance-criteria gate cannot distinguish "adding the first AC" from "rewriting AC after implementation"; the documented override for it reports success and writes nothing; and `--output silent` suppresses the rejection while returning exit 0. Together, a mutation can disappear with no error, no output, and a success exit code.

## 1. An empty task cannot be given its FIRST acceptance criteria (exit 48)
```
cleo update T236 --acceptance "..."
# -> Acceptance criteria locked at stage implementation.
#    Reframing AC after implementation is anti-pattern. (48)
```
**T236 has ZERO acceptance criteria and no description.** The gate exists to prevent moving the goalposts after implementation, but it cannot tell that case from establishing a bar that was never set. Consequence: a task an agent is **actively building against right now** is frozen with no acceptance criteria at all — the precise opposite of the gate's purpose. `--acceptance` is documented as REQUIRED for all task creation (ADR-066), so a task that reached `implementation` with none is already anomalous; the gate then makes it permanently unfixable.

**Ask:** when the existing AC list is empty, treat the write as establishing criteria rather than reframing them, and allow it.

## 2. The `--reason` override is accepted, exits 0, and writes NOTHING
```
cleo update T236 --acceptance "..." --reason "establishing first AC on an empty task"
# -> exit 0, no error
cleo show T236 --full
# -> description: empty, acceptance: empty
```
**This is the serious one.** The documented escape hatch for (1) reports success and silently discards the write. An agent following the documented path has no way to learn its write did not happen except by re-reading — and re-reading is exactly what agents are told is redundant because mutation envelopes are self-confirming.

**Ask:** either honour `--reason` or reject it with a non-zero exit and a named error. Reporting success for a discarded write is worse than refusing.

## 3. `--output silent` makes a REJECTED write indistinguishable from a successful one
With `--output silent`, the exit-48 rejection in (1) prints nothing **and `$?` is 0**. The flag is documented as "Silent (exit-code only)", which is precisely the contract it breaks: the exit code is the only channel left and it does not carry the failure.

**Ask:** `--output silent` must preserve a non-zero exit on a failed mutation. This is the flag most likely to be used in scripted/batch agent work, so it is the worst place for a swallowed error.

[...]

PR #1279, opened 2026-09-12
fix(T12142): keep stdout to one LAFS envelope — route AI SDK warnings to stderr (#1223)

kryptobaseddev:
Closes GH #1223 (the reproducible half).

ADR-086 states stdout is ONE LAFS envelope per call and all logs go to stderr. On `cleo update` it was not:

```
stdout line 0: {"success":true,"data":{"count":1,"updated":["T12142"],…
stdout line 1: AI SDK Warning System: To turn off warning logging, set the
               AI_SDK_LOG_WARNINGS global to false.

  -> json.loads(stdout): Extra data: line 2 column 1 (char 744)
```

The banner lands **after** the envelope, so parsing the whole of stdout as JSON fails outright. Any consumer parsing stdout — which the `--field` contract actively encourages — has to skip lines first, which is the exact "pipe through `tail`/`jq`" anti-pattern ADR-086 forbids.

## Mechanism

`ai@6`'s `logWarnings` splits its output across both streams, and puts the wrong half on stdout:

```js
// node_modules/ai/dist/index.mjs
console.info(FIRST_WARNING_INFO_MESSAGE);   // stdout  <- breaks the contract
for (const warning of options.warnings) {
  console.warn(formatWarning({ … }));       // stderr  <- correct
}
```

`console.info` is stdout in Node. So the SDK's *banner* polluted the envelope while its *warnings* were already going to the right place.

## Scope: mutating commands only

All 13 read commands swept (`show`, `list`, `find`, `current`, `version`, `session status`, `saga list`, `labels`, `blockers`, `docs list-types`, `memory llm-status`, `backup list`, `next`) emit exactly one envelope and parse clean. The pollution appears on the write path, where the LLM enrichment runs.

## Fixed with a handler, not by silencing

`globalThis.AI_SDK_LOG_WARNINGS` accepts `false` **or a function**. Setting `false` would satisfy the contract by destroying the information, and these warnings are real: the one that surfaced this said `responseFormat is not supported` for `ollama.chat/qwen2.5-coder:3b` — exactly what you want to know when a local model silently ignores a JSON-schema request.

So the warnings are kept and routed to `getLogger('llm:ai-sdk').warn`. Installed at module load of `llm/model-runner.ts`, which AGENTS.md gate 13 makes the SSoT for LLM client construction — the one place every consumer (CLI, studio, daemon) passes through. Idempotent, and it never overwrites a handler an embedder already set or an explicit `false`.

## Why no existing gate caught it

`lint-stdout-discipline` and `lint-stdout-write-allowlist` scan **our** source. This was a dependency's runtime `console.info`, which no static analysis of this repo can see. That is a real gap in the gate family: **stdout purity needs a runtime assertion, not only a static one.** The tests here assert it at runtime by spying on `console.info`, `console.log` and `process.stdout.write`.

## Not in this change

#1223 reports two other stdout lines, neither of which I could attribute:

- **`[LocalBackend] @vlcn.io/crsqlite not installed …`** — not reproducible here, and `grep -rn LocalBackend` over `packages/` finds only a TSDoc mention in `contracts/src/attachment.ts`. It is llmtxt's `LocalBackend`, reached only on a doc/attachment path, and almost certainly the same `console.info`/`console.log` class.
- **`mergeError: task branch 'task/T1655' does not exist`** — this one is **not** stdout pollution. It goes through `getLogger` (`core/src/tasks/complete.ts:1145`), which routes to a log *file* when initialised and to stderr otherwise. The reporter's substantive point still stands though: it is logged at WARN for tasks that never had a task branch, where nothing actually failed, so it reads as a failure and is not one. That is a one-line condition in `complete.ts` and is tracked separately.


kryptobaseddev:
## End-to-end verification against the built artifact

Unit tests prove the handler behaves; they do not prove it is *installed*, or that a real warning survives the trip. Both checked against `packages/core/dist/`:

```
before import: undefined
after import : function          <- module-load install fires in the SHIPPED build

{"level":"WARN","subsystem":"llm:ai-sdk","provider":"ollama.chat",
 "model":"qwen2.5-coder:3b","warning":"responseFormat is not supported",
 "msg":"AI SDK warning"}          <- information preserved, with provider + model

stdout bytes written by the handler: 0
```

So the warning is not merely moved off stdout — it arrives **structured**, carrying the provider and model, which the SDK's own `console.warn` string did not.

## stdout purity on a real mutation, three consecutive runs

```
run1 exit=0 | stdout: PURE 704 bytes | run2 exit=0 | PURE 704 bytes | run3 exit=0 | PURE 704 bytes
```

Before the fix the same command produced **841 bytes** that failed to parse — `Extra data: line 2 column 1 (char 744)`. The 744 boundary is the envelope; the remaining 97 bytes were the banner.

## One measurement that looked like a regression and was not

My first post-fix run of `cleo update` returned **exit 124** — `timeout` killing it at 400s — where the pre-fix run had completed. That reads exactly like an introduced hang, so I measured rather than assumed:

```
current  run1: exit=0  2s      current run2: exit=0  2s
update        : exit=0  3s      (and exit=0 on three further runs)
```

The same command completes in 3 seconds. The 124 was the pre-existing intermittent write-path hang — the one separately reported as `cleo update` blocking past 120s (#1235 item 4) and root-caused as teardown awaiting a never-settling step. Not caused by this change.

Recording it because an intermittent pre-existing hang is precisely the kind of thing that gets misattributed to whatever diff happens to be in the tree — and the honest resolution was three more runs, not an argument.

kryptobaseddev:
## Root cause, at the source: the AI SDK splits its two messages across two streams, and the one on stdout is the meta-banner

This PR was opened on the reasonable assumption that AI SDK warnings were noise on the wrong
stream. The defect is narrower and worse than that: **the actual warning is correctly on
stderr. The banner telling you how to turn warnings off is on stdout**, and it is the one
that corrupts the envelope.

`node_modules/ai/src/logger/log-warnings.ts` (ai@6.0.162, the version in the installed
v2026.8.9 dist):

```ts
export const FIRST_WARNING_INFO_MESSAGE =
  'AI SDK Warning System: To turn off warning logging, set the AI_SDK_LOG_WARNINGS global to false.';

let hasLoggedBefore = false;

export const logWarnings: LogWarningsFunction = options => {
  if (options.warnings.length === 0) return;
  const logger = globalThis.AI_SDK_LOG_WARNINGS;
  if (logger === false) return;
  if (typeof logger === 'function') { logger(options); return; }

  // display information note on first call
  if (!hasLoggedBefore) {
    hasLoggedBefore = true;
    console.info(FIRST_WARNING_INFO_MESSAGE);   // ← stdout
  }

  for (const warning of options.warnings) {
    console.warn(formatWarning({ warning, ... }));   // ← stderr
  }
};
```

Stream behaviour measured on this host (node v24.16.0), rather than assumed:

```
$ node -e 'console.info("INFO-LINE")' >o 2>e     →  stdout: INFO-LINE   stderr: (empty)
$ node -e 'console.warn("WARN-LINE")' >o 2>e     →  stdout: (empty)     stderr: WARN-LINE
```

So `console.info` → **stdout**, `console.warn` → **stderr**. The SDK puts the warning on the
right stream and the *instructions for silencing it* on the wrong one.

## Why this is a contract break, not a cosmetic banner

Captured by cleo-dev while creating task rows — note `2>/dev/null`, which discards the
`console.warn` half and leaves the `console.info` half:

```
ID2=$(timeout 240 cleo add … --field /data/created/0 2>/dev/null)
echo "ID2=$ID2"
ID2=T12162
AI SDK Warning System: To turn off warning logging, set the AI_SDK_LOG_WARNINGS global to false.
```

`--field /data/created/0` is the scalar-extraction path ADR-086 documents and
CLEO-INJECTION.md instructs **every** agent to use for create-then-use. It returned
`"T12162\nAI SDK Warning System: …"`.

That is a **corrupted task id handed back by the documented create path**, and it is another
manufacturing route for the wrong-task-id problem that has hit three sessions eight times in
one day (see #1244, #1302). That instance survived only because unquoted word-splitting
dropped the remainder and the next command ignored the extra args — it worked by accident.

**`hasLoggedBefore` is module-level state, so it fires once per PROCESS.** Every `cleo`
invocation is a fresh process, so this is not a one-time nuisance on first use — it is **every
invocation that triggers a warning**.

## This PR's fix is complete, and the source says why

`installAiSdkWarningHandler()` sets `globalThis.AI_SDK_LOG_WARNINGS` to a function. Read the
control flow above: the `typeof logger === 'function'` branch **returns before**
`hasLoggedBefore` is ever consulted. So assigning a function suppresses the `console.info`
banner as well as redirecting the warnings — both halves, not just the one this PR was aimed
at. No additional change is needed here; the fix is broader than its original rationale.

Two things worth adding to this PR's description on that basis:

1. The regression test should assert **stdout purity on a write command that emits a
   warning** — specifically that `--field /data/created/0` yields exactly one line — not just
   that warnings appear on stderr. The stderr assertion passes even with the banner leaking.
2. `AI_SDK_LOG_WARNINGS = false` would also suppress both, but a function is the right choice:
   it preserves the warnings on stderr where an operator can see them, rather than discarding
   diagnostic information to fix a stream-routing bug.

## Adjacent findings from the same probe (cleo-dev)

**The banner fires on `cleo add` and not on `cleo show` / `cleo find`** — both checked, stdout
clean, stderr empty. An AI SDK client is therefore being constructed on the **write** path
specifically. That is direct evidence for the open question on #1244 — *does `cleo add` need
an LLM at all?* — and it means a task create is blockable by a model load.

**`cleo add-batch --dry-run` exceeded 120 s on a single-task file**, and a second run of the
same probe has been going several minutes at time of writing. A dry run is the operation whose
entire contract is to mutate nothing; spending minutes of model work before deciding not to
write corroborates the O(active-tasks) embedding-before-insert path that #1258 bounds.

Both belong to the LLM-on-the-write-path question rather than to stream routing, so they are
recorded here as context and are not fixed by this PR.

kryptobaseddev:
## Measured on a compiled CLI: all three output modes are clean, and the LLM genuinely ran

The earlier analysis here was source-level. It has now been run against a real build
(cleo-optimizer's full workspace build), with a write command that actually triggers an AI SDK
warning and the two streams captured to separate files:

| mode | exit | stdout bytes | banner on stdout | banner on stderr |
|---|---:|---:|---:|---:|
| `--field /data/count` | 124 | 2 | **0** | 0 |
| `--output id` | 124 | 8 | **0** | 0 |
| `--output silent` | 0 | **0** | **0** | 0 |

**This is not a "the model never ran" non-result.** stderr in all three carries the handler's
output:

```json
{"level":"WARN","subsystem":"llm:ai-sdk","provider":"ollama.chat",
 "model":"qwen2.5-coder:3b",
 "warning":{"type":"unsupported","feature":"responseFormat",
            "details":"JSON response format schema is only supported with structuredOutputs"},
 "msg":"AI S…"}
```

The model was invoked, the warning fired, and it was routed through the installed handler —
so the measurement exercises the defect path rather than avoiding it. Zero bytes reached
stdout in any mode, **including `--output silent`**, which was the one worth checking: a
banner there would have made "silent" false in the strongest possible way.

It is also an improvement over what it replaces. The upstream default is a bare
`console.warn` string; this is structured JSON carrying `subsystem`, `provider`, `model` and
the typed warning — parseable by an operator rather than only readable.

### Attribution limit, stated rather than glossed

This build contains **both** install sites: the new one in `runMainWithLafsEnvelope` and the
retained one at `model-runner.ts` module load. `provider: "ollama.chat"` does not distinguish
them, so this measurement establishes **that the product is clean**, not **that the funnel
install is the site doing the work**.

What is established independently, by reading rather than running, is the *coverage* argument:
`memory/llm-backend-resolver.ts` constructs its own clients and imports `ai` only as
`import type { LanguageModel }` — type-only, erased at runtime — and does not import
`model-runner.js`. So that path cannot install the handler via the chokepoint. Whether some
*other* module in a given invocation happens to load `model-runner` first is a separate
question, and one this run cannot answer.

A build with the funnel install removed would settle it, and is being run. If the banner leaks
without the funnel install, the relocation is load-bearing as claimed. **If it does not leak,
the original placement was accidentally sufficient in this configuration** — which would not
make the relocation wrong (the coverage gap is real and structural), but would mean the commit
message overstates its urgency, and that should be corrected rather than left standing.

kryptobaseddev:
## Correction: my commit message overstated this, and the measurement that shows it is now in

I asked for a build with **this PR's install removed** specifically to test my own
justification, and it came back the way I did not predict. Amended in `a7198c216`; the code
is unchanged, only the claim.

### The experiment

Funnel install removed, chokepoint install intact, write command that genuinely invokes the
model:

```
src/cli/index.ts   installAiSdkWarningHandler: 0   (removed)
model-runner.ts                              : 2   (intact)

exit=124   stdout=2B   banner_on_stdout=0   stderr=313B   llm_ran=1
```

**stdout stayed clean without this PR's change.** So the emission is already covered today —
some other module on that path pulls `model-runner.ts` in incidentally, and installing the
handler is a side effect of that import. **The original placement was accidentally
sufficient.**

The artifact was verified rather than assumed, after a grep-for-an-identifier gave a false
negative earlier in the day: presence was checked with a **string literal**
(`grep -c "llm/ai-sdk-warnings"`), which survives compilation where an identifier does not,
and dist mtime confirmed a rebuild 2s after the edit.

### What the commit now claims instead

The earlier message implied an **active leak** on the resolver path. That was wrong, and the
kind of wrong worth correcting explicitly rather than quietly: it would have had a reader
believe a measurement that had not been taken.

The justification is **structural, not empirical**:

- `llm-backend-resolver.ts` constructs its own AI SDK clients and imports `ai` only as
  `import type { LanguageModel }` — type-only, erased at runtime — and does not import
  `model-runner.js`. That is verified by reading, independently, twice.
- So nothing *on that path* installs the handler. Today's safety comes from an unrelated
  module happening to import the chokepoint.
- That is a fact **nobody wrote down, no test asserts, and a lazy-import optimisation or a
  dependency reshuffle would remove silently.**

A guard that works because of an incidental import is the same shape as an invariant asserted
in a comment with nothing enforcing it. Moving the install to the funnel removes the
dependency on that accident; it does not fix an outage that was happening.

### Why this makes the change easier to review, not harder

*"We moved it and it kept working"* is a weak claim. *"We moved it off a guarantee that was
never real and happened to hold"* is a specific one, and it is falsifiable — the placement
tests in this PR assert the install sits in the funnel and that the bypassing module still
does not import the chokepoint, so the property is now pinned rather than inherited.

### Unchanged and still measured

All three extraction modes are clean on the shipped configuration, with the model genuinely
invoked (structured `llm:ai-sdk` warning present on stderr in every run):

| mode | exit | stdout | banner on stdout |
|---|---:|---:|---:|
| `--field /data/count` | 124 | 2 B | 0 |
| `--output id` | 124 | 8 B | 0 |
| `--output silent` | 0 | 0 B | 0 |

Credit to cleo-optimizer for running both builds, including the one designed to contradict me.

PR #1309, opened 2026-09-12
fix(T12161): cleo doctor warns when a shared .git leaks stash or identity between sessions

kryptobaseddev:
A `.git` shared by several worktrees also shares two things that look tree-local and are not. Both bit this repository on 2026-09-12, in separate incidents hours apart, and nothing warned in either case.

## The stash stack is repository-wide

`git stash pop` with no argument takes `stash@{0}` — whichever entry was pushed most recently by **any** worktree. An agent ran `git stash push` on an untracked file (a no-op, so nothing of its own was pushed) then `git stash pop`, and silently received an unrelated entry from a 26-deep stack belonging to another session's branch. It modified a generated manifest the agent had never touched, and was caught by reading the diff before committing — not by tooling.

Measured here now: **7 worktrees, 27 stash entries**, `stash@{0}` belonging to `task/T11185`. Entries date back months, across branches long since merged.

## The committing identity lives in that same shared config

`user.name` / `user.email` set at `--local` scope re-author every commit made by every other session in every other worktree. A `compose probe <probe@local>` identity left behind by a merge-composition experiment authored three commits from a different session on a different branch before anyone noticed.

## Design notes

Both checks are **gated on the `.git` actually being shared**. A stash in a single-worktree repo is private; a local identity there affects only the person who set it. Warning about either would be noise.

The identity check reads `--local` rather than the effective identity, and **that distinction came from a test failing honestly**. The first version read `git config user.email`, which returns the MERGED value and falls through to the committer's global config — so it flagged any first-time contributor whose correct global identity simply had no commits here yet. A repo-local override is different in kind: it is the only sort one session can impose on another through the shared `.git`. So the remedy is to REMOVE the override and fall back to the committer's own identity, not to set another value. The regression is pinned as its own test.

Neither check mutates anything and neither fails the run: both report `warning`, because a shared stash and a deliberate identity override are both legitimate. **The defect is that they are invisible, not that they are wrong.**

## Verification

- 11/11 unit tests against **real repositories with real worktrees and real stash entries**. Git's cross-worktree sharing semantics are the behaviour under test, so a mock would assert our belief about git rather than git — and that belief is what was wrong in the first place.
- A/B typecheck: 120 errors with the change, 120 without (all pre-existing unbuilt-workspace-dep noise), delta 0.
- Live behaviour confirmed: stash check fires (7 worktrees / 27 entries), identity check passes now that the override matches history.


kryptobaseddev:
## Live instance: this check fires on the current repo, 27 entries across 8 worktrees

Measured on `/mnt/projects/cleocode` at `main`, as the first real-world case rather than a
constructed one. The check's firing condition is `entries.length > 0 && worktrees >= 2`; the
live inputs it reads:

```
.git at repo root      : exists (directory)
stash entries          : 27
worktrees sharing .git : 8
→ condition FIRES — 27 entries across 8 worktrees
top entry reported     : stash@{0}: WIP on task/T11185: 9340d712a T11179
```

So the check is not vacuous — it triggers today, on the repo it was written for, without
anyone constructing the condition.

### What is actually in there

```
entries     : 27
content     : 11,912 insertions total
dates       : ALL 2026-05-26 / 2026-05-27 — three and a half months old
largest     : 28 files / 2,301 insertions  (stash@{3}, "WIP on main: T11134 spec")
entries referencing the T121xx task range in use today : 0
```

That last line is the load-bearing one: it establishes the stack is **not** current-session
work without anyone having to inspect 27 diffs. Every message names `task/T109xx`–`task/T111xx`
branches from May.

### The justification this check deserves, beyond "someone might pop the wrong one"

A stash is invisible to **all four** of the commands this fleet has been treating as proof of
"nothing outstanding":

```
git log --branches --not --remotes    ← sees no stash
git branch                            ← sees no stash
git worktree list                     ← sees no stash
gh pr list                            ← sees no stash
```

**Four independent instruments agreeing on an absence, none of which could have produced a
positive.** That is the same shape as every other absence defect catalogued this week, and it
is a stronger argument for the check than the pop-collision that prompted it: the collision is
an accident someone might avoid by care, this is a blind spot no amount of care closes, because
the reader has no signal that a blind spot exists.

### The mechanism that makes it unresolvable by care

**A linked worktree shares the stash stack with its parent.** `git stash list` run inside a
worktree returns the parent checkout's entries — verified here, my worktree returns the
identical 27.

So an agent checking "is there anything of mine outstanding?" in their own worktree sees 27
stashes and has **no way to tell** whether they are their own forgotten work or another
session's. Both readings are available and the command distinguishes neither. Gating the check
on `>1 worktree sharing the .git` targets exactly that condition.

### Disposition

Reported, not acted on. `git stash drop` is irreversible and 11,912 insertions of
three-and-a-half-month-old work is the owner's call, not an agent's.

kryptobaseddev:
## Correction: my "11,912 insertions" is size-when-made, not work-outstanding

My earlier comment reported the stash stack as **27 entries / 11,912 insertions** in a context
that invited reading it as outstanding work. **That framing is wrong and I am withdrawing it.**

`git stash show --stat` diffs a stash against **its own parent**, not against current `main`. It
reports how big the change was *when it was made*. For entries from 2026-05-26/27 that number
says nothing about whether the work ever landed — and a stash whose content was subsequently
merged reports exactly the same insertion count as one that was never touched again.

The credit is axiom-dev-lead's: they used the same method on three stashes in a different repo,
reported one as holding a production race, then re-measured per-file against main and found
**all three already landed**. The correct test is per-file `git diff <stash-sha> origin/main -- <path>`.

### Re-measured properly, all 27 entries, 209 files

```
files total            : 209
IDENTICAL to main      :  21   ← conclusively landed
differ from main       : 188   ← INCONCLUSIVE, see below
```

**A nonzero diff is not evidence of outstanding work.** These entries are three and a half
months old and `main` has moved substantially since; a file differing from today's `main` is the
expected result whether the stashed change landed or not. Only the zero case is conclusive.

So the honest statement is: **21 files are provably landed, and the other 188 are unassessed.**
I cannot say from this measurement that the stack holds unlanded work, and I should not have
implied a number that reads as though it does.

### What survives, unchanged

The reason this check is worth running is untouched, and it is not about volume:

> A stash is invisible to `git log --branches --not --remotes`, `git branch`,
> `git worktree list` **and** `gh pr list` — four independent instruments agreeing on an
> absence, none of which could have produced a positive.

And the mechanism that makes it unresolvable by care is untouched: **a linked worktree shares
the stash stack with its parent**, so `git stash list` inside a worktree returns the parent
checkout's entries, and an agent can read them as their own forgotten work or dismiss them as
someone else's with no way to distinguish.

**"Invisible" and "outstanding" are different claims.** This check establishes the first. It
was my error to present a size figure as though it established the second.

### Two refinements worth carrying into the check's output

1. **Report the per-file diff against the default branch, not `--stat`.** A `stat` figure will be
   read as outstanding work every time — I did it, and so did the person who warned me about it
   the same day. If the check emits a number, that number should be the one that means something.
2. **Cite the stash SHA, never the index.** `stash@{n}` renumbers whenever any entry is dropped —
   axiom-dev-lead cited the same stash as `stash@{2}` and later `stash@{1}` within one session,
   having warned someone else about exactly that. The SHAs are in my re-measurement above.

PR #1312, opened 2026-09-12
fix(T12163): pretypecheck/prebuild verify the manifest, never regenerate it (gh#1306)

### LINKED ISSUE #1306: [Bug]: pretypecheck/prebuild regenerate a TRACKED file, so running typecheck invalidates the evidence cache for every tool
Second instance of the gh#1221 defect — "a tool invalidates its own cache by running" — in a different place, and the fix for the first one does not reach it.

### Measured

`packages/cleo/package.json`:

```json
"gen:manifest":  "node scripts/generate-command-manifest.mjs",
"prebuild":      "node scripts/generate-command-manifest.mjs",
"pretypecheck":  "node scripts/generate-command-manifest.mjs",
```

So every `pnpm build` and every `pnpm typecheck` — local or CI — regenerates `packages/cleo/src/cli/generated/command-manifest.ts`, which is **tracked**.

```
$ git status --porcelain=v1 --untracked-files=no | wc -l
0
$ pnpm --filter @cleocode/cleo run gen:manifest
$ git status --porcelain=v1 --untracked-files=no | wc -l
1
 M packages/cleo/src/cli/generated/command-manifest.ts
```

### Why that breaks evidence caching

`captureDirtyFingerprint` (`packages/core/src/tasks/tool-cache.ts`) runs `git status --porcelain=v1` over the repo with `.cleo/` sub-paths excluded. `packages/cleo/src/cli/generated/` is not excluded, so the regenerated file enters the fingerprint.

The evidence cache is keyed on `(canonical, cmd, args, HEAD, dirtyFingerprint)`. Therefore:

- `cleo verify --evidence "tool:typecheck"` runs typecheck → `pretypecheck` dirties a tracked file → **the fingerprint changes as a side effect of the run itself**.
- Any subsequent `tool:test`, `tool:lint` or `tool:typecheck` sees a different fingerprint and misses cache.
- The tools invalidate each other's entries, indefinitely, on a path that fires every time.

### Why the gh#1221 fix does not cover it

That one was about **untracked** noise, and the remedy was narrowing the fingerprint capture to `--untracked-files=no`. This mutation is to a **tracked** file, so the narrowing is irrelevant here — and in any case that change is still unmerged, so neither half is on `main` today.

The two cases share a shape and need different remedies. It is worth stating the general rule: **a build step that mutates tracked files is an evidence-fingerprint hazard**, and the fingerprint cannot defend itself against it.

### Second-order cost

Every agent who runs typecheck or build gets a dirty tree they did not create. Measured today: a maintainer ran `pnpm --filter @cleocode/cleo run typecheck` and got a 201-line modification to a file they never touched, and reasonably suspected another session's stash had leaked into their worktree. The real cause is a pre-hook.

Most of that 201 lines is a generator-versus-biome formatting race — the generator emits unformatted, biome formats, the formatted version is committed, the generator un-formats it. That noise matters because it **camouflages real drift**: after formatting the regenerated output, 18 lines of genuine content difference remain, including two commands present in source and missing from the committed artifact (`cleo doctor memory-guard`, `cleo doctor superseded-store`).

### Suggested fix

1. **`pretypecheck` should VERIFY, not regenerate.** Generation belongs in an explicit `gen:manifest` step and in CI's drift gate; a typecheck should be read-only with respect to tracked files. This is the change that removes the hazard rather than reducing it.
2. **Make the generator format its own output**, so a diff means content drift and nothing else. Independently worth doing — it is why the two missing commands went unnoticed.
3. Consider whether any other `pre*` hook mutates tracked files; the same reasoning applies to each.

### Not claimed

No runtime breakage. `memory-guard` and `superseded-store` both run fine (`exit=0`) because they are `doctor` subcommands registered directly in `doctor.ts`, not top-level manifest entries. The staleness is real but its live impact is on the cache and on tree cleanliness, not on dispatch.

### Related

- gh#1221 — the first instance (untracked noise).
- gh#1281 — generated-and-committed files desyncing on merge; PR #1282 adds the drift gate that **does** catch this artifact's staleness (verified: run against `main` it fails with exactly the 18-insertion diff).

kryptobaseddev:
Closes #1306. Ordered deliberately **before** #1282's manifest commit — see the last section.

## A build step was mutating a tracked file

`pretypecheck` and `prebuild` both ran the generator, so every typecheck and every build rewrote `packages/cleo/src/cli/generated/command-manifest.ts`:

```
$ git status --porcelain=v1 --untracked-files=no | wc -l
0
$ pnpm --filter @cleocode/cleo run gen:manifest
$ git status --porcelain=v1 --untracked-files=no | wc -l
1
 M packages/cleo/src/cli/generated/command-manifest.ts
```

**Two costs.** Every agent got a dirty tree they did not create — one maintainer reasonably suspected another session's stash had leaked into their worktree. And `captureDirtyFingerprint` feeds `git status` into the evidence cache key `(canonical, cmd, args, HEAD, dirtyFingerprint)`, so **`tool:typecheck` changed the fingerprint as a side effect of running**, and every subsequent `tool:test` / `tool:lint` / `tool:typecheck` missed cache. The tools invalidated each other indefinitely, on a path that fires every time.

This is the **second instance of #1221's shape** — *a tool invalidates its own cache by running* — and the first remedy does not reach it. That one was about **untracked** noise (`--untracked-files=no`); this is a **tracked** mutation. Worth stating generally:

> A build step that mutates tracked files is an evidence-fingerprint hazard, and the fingerprint cannot defend itself against one.

`--check` verifies and never writes. `gen:manifest` remains the explicit write path. Verified: with `--check` wired in, a typecheck leaves the tracked-file count unchanged.

## The generator now formats its own output, and that is what makes `--check` meaningful

It previously emitted unformatted source; biome reformatted it; the formatted version was committed; the next run un-formatted it again — roughly **183 lines of pure formatting churn per regeneration**.

That noise was not cosmetic. **It camouflaged real drift.** Two commands were missing from the committed manifest inside it:

```
+    name: 'memory-guard',
+    name: 'superseded-store',
```

`cleo doctor memory-guard` is the documented remedy for the P0 machine-freeze issue. `cleo doctor superseded-store` is what AGENTS.md prescribes for the which-DB-is-real confusion. Both are commands AGENTS.md actively tells agents to reach for, and both sat missing inside formatting noise.

The same regeneration now produces **18 insertions and 1 deletion — pure content** — where it previously produced 201 lines.

## One measured subtlety, recorded at the call site

The formatting scratch file **must live in the output directory**, not `os.tmpdir()`.

Biome resolves configuration and per-path overrides **by path**, so formatting identical bytes at `/tmp` and at `packages/cleo/src/cli/generated/` yields different output. I caught it with a flip-flop check — the temp-path result was immediately reformatted at the real path:

```
$ node generate-command-manifest.mjs && cp out /tmp/before
$ pnpm biome format --write <real path>
Files /tmp/before and <real path> differ     ← would make --check fail forever
```

After moving the scratch file into the output directory:

```
Formatted 1 file in 8ms. No fixes applied.
STABLE — no flip-flop
generate-command-manifest: … is current (163 entries).
```

Worth having in the source, because the naive fix is the temp-file one and it fails in a way that only shows up after the commit.

## Ordering

The regenerated manifest is committed here so the tree is self-consistent when this lands, and this must land **before** #1282 commits its own copy. If the regenerate-on-typecheck behaviour were still live when a canonical manifest landed, every agent's next typecheck would re-dirty the file we had just declared canonical — a file that rewrites itself.

## Not claimed

**No runtime behaviour changes.** `memory-guard` and `superseded-store` always worked, because they are `doctor` subcommands registered directly in `doctor.ts` rather than top-level manifest entries. The staleness cost was to the cache and to tree cleanliness, not to dispatch.


kryptobaseddev:
## The defect demonstrated in the wild, an hour after filing it

While working on an unrelated branch (#1279's, which does not carry this fix), I ran a full workspace build so a spawned-CLI test could reach the compiled binary. Afterwards:

```
$ git status --short
 M packages/cleo/src/cli/generated/command-manifest.ts

$ git diff --stat
 packages/cleo/src/cli/generated/command-manifest.ts | 201 ++++++++-------------
 1 file changed, 75 insertions(+), 126 deletions(-)
```

A clean checkout. A build. A 201-line modification to a generated file nobody touched, left behind in the working tree.

I had not edited that file, I was not working on the manifest, and I was on a different branch for a different issue. The only reason I recognised it rather than investigating it was that I had filed #1306 an hour earlier — which is exactly the cost the issue describes: a maintainer hitting this cold reasonably suspects another session's stash leaked into their worktree, because that is a far more plausible explanation than "a read-only-sounding command rewrote a tracked file".

Two things this adds to the case in the PR body:

1. **It is a build as well as a typecheck.** The issue's reproduction used `gen:manifest` directly; this is `pnpm -r build` going through `prebuild`, in normal use, with no intent to touch the manifest.
2. **The 201 lines are the camouflage argument, concretely.** On this branch — generator un-formatted, biome-formatted file committed — regeneration produces 201 lines of churn. With this PR's self-formatting, the same regeneration produces **18 insertions and 1 deletion, all content**. The two missing doctor commands were invisible inside the first number and unmissable in the second.

Also worth recording for whoever reviews the fingerprint interaction: the tracked-file mutation is what makes this an evidence-cache problem rather than a tidiness one. `captureDirtyFingerprint` runs `git status --porcelain=v1` with `.cleo/` sub-paths excluded; `packages/cleo/src/cli/generated/` is not excluded. So the build above changed the evidence cache key for every subsequent `tool:*` verification in that tree, as a side effect of building.

Issue #1329, opened 2026-09-12
[Bug]: cleo nexus status reports indexed:true with a full node count for any path, including one that does not exist


kryptobaseddev:
Found incidentally while measuring #1293. Not the issue I was working on, and worth its own report because the answer is confidently wrong rather than absent.

### Measured

```
$ cleo nexus status /definitely/not/a/real/repo --json
success=true  indexed=true  nodes=26964
repoPath=/definitely/not/a/real/repo
projectId=L2RlZmluaXRlbHkvbm90L2EvcmVhbC9y
```

A path that does not exist reports **`indexed: true`** with the **full node count of the real project**, and `success: true`.

The `projectId` is derived correctly from the bogus path — `base64url("/definitely/not/a/real/r…")` — so the command *knows* it is being asked about a different project. It then reports counts that are not scoped to it.

### The same defect reached by a second route

I arrived here because a flag value got swallowed as a positional:

```
$ cleo nexus status --output envelope --json
repoPath  = /mnt/projects/cleocode/envelope
projectId = L21udC9wcm9qZWN0cy9jbGVvY29kZS9l   → "/mnt/projects/cleocode/e"
indexed   = true      nodes = 26964

$ cleo nexus status --json                      # control
repoPath  = /mnt/projects/cleocode
projectId = L21udC9wcm9qZWN0cy9jbGVvY29kZQ     → "/mnt/projects/cleocode"
indexed   = true      nodes = 26964
```

**Two different project identities, byte-identical status.** So `--output` is not a recognised flag on this subcommand and citty takes its value positionally — which silently redirects the query to a nonexistent project — and the status surface then fails to notice.

That second half is the part I would fix first. A swallowed flag is a parsing bug; a status that cannot distinguish a real project from an invented one is a **reporting** bug, and it is the one that misleads.

### Why it matters more than it looks

`CLEO-INJECTION.md` makes this command the mandated first call for the whole subsystem:

> **FIRST CALL IS `cleo nexus status`.** It returns `nodeCount`, `lastIndexedAt` and `staleFileCount`. The index is NOT auto-refreshed — a symbol added after the last `analyze` run is simply absent, and a lookup for it returns `E_NOT_FOUND` that reads identically to "no such symbol".

[...]

Issue #1278, opened 2026-09-12
[Bug]:The critical-gate refusal proposes the bypass as its own remedy: it suggests files: evidence for testsPassed


kryptobaseddev:
## The defect

`CLEO_OWNER_OVERRIDE` correctly refuses to cover critical gates — `E_CRITICAL_GATE_OVERRIDE_REJECTED` (T9245). That refusal is right and should stay.

**But the refusal text then suggests `commit:<sha>;files:<paths>` as the way forward — and CLEO accepts a `files:` atom for `testsPassed`.**

So the remedy offered for *"you may not self-attest that tests passed"* is *"cite some file paths as your test evidence"*. The guard blocks the honest override and recommends the dishonest one.

## Why this matters more than an awkward message

The whole point of ADR-051 is that a gate write must be backed by evidence CLEO can validate against git, the filesystem or the toolchain. A `files:` atom proves a file exists with a given hash. **It proves nothing whatsoever about whether a test ran, let alone passed.** Accepting it for `testsPassed` makes that gate satisfiable by any agent that can name a path — which is every agent.

And the refusal message is the worst possible place to put that suggestion, because it arrives at exactly the moment an agent is looking for a sanctioned way through a blocked gate. A guard that refuses a bypass and then names a weaker one in the same breath trains the fleet toward the weaker one. The agent is not circumventing anything — it is following the error message.

This interacts badly with a condition that is live on at least two projects here: **`pr:<n>` atoms cannot validate at all** when a repo has no CI (GitHub Actions removed; required checks empty), so `testsPassed` has no legitimate satisfier left. `tool:test` is the only honest one, and on a memory-constrained box it is the most expensive thing an agent can run. That combination — no `pr:`, expensive `tool:test`, and a refusal message pointing at `files:` — makes the weak path the path of least resistance precisely where the gate matters most.

## Suggested fix

1. **Reject `files:` for `testsPassed`.** It is not test evidence. Accept only `tool:test` or `test-run:<json>`.
2. **Fix the refusal text** to name only atoms that are valid for the gate being refused. An error message is a recommendation; it should not recommend something the validator will accept but that proves nothing.
3. Consider stating in the refusal *why* the gate is critical, rather than offering an alternative route at all. "This gate requires an executed test run; there is no override" is more useful than a suggestion that undermines it.

## Provenance

[...]

Issue #1245, opened 2026-09-12
[Bug]:cleo list --severity is silently IGNORED: P0, P3 and an invalid value all return every task


kryptobaseddev:
## The defect

`cleo list --severity <Px>` does not filter. Every value returns the entire task set, and the outputs are byte-identical.

Measured 2026-09-12 on a project with 291 tasks:

```
cleo list --output count                 -> 291    (total)
cleo list --severity P0 --output count   -> 291
cleo list --severity P3 --output count   -> 291
cleo list --severity BOGUS --output count -> 291    <-- invalid value, no error
cleo list --status done --output count   ->  63    (--status works correctly)
```

`md5sum` of `--severity P0 --output table` and `--severity P3 --output table` are **identical**.

So the flag is parsed and accepted, then discarded.

## Why this is Major, and why it is worse than "a filter doesn't work"

**It fails OPEN, and in the direction that produces a confident wrong answer.**

An agent asked to account for the P0 work runs `cleo list --severity P0`, receives **every task in the project**, and has no signal that the filter did nothing. The natural reading of that result is *"these 291 tasks are all P0"* — the opposite of the truth for ~290 of them. There is no error, no warning, and the envelope looks exactly like a successful narrow query.

This is the same shape as a guard that fails open on an absent field: the absence of filtering is indistinguishable from "nothing was filtered out". Compare `--status`, which works — so a user has every reason to believe `--severity` does too, and the two are used interchangeably in reconciliation work.

**`--severity BOGUS` not erroring is the compounding half.** A rejected value would have exposed the whole defect instantly: anyone typo-ing a severity would have seen it. Instead an unvalidated enum silently widens the query to everything.

`--severity` is documented as a first-class axis (ADR-066: *"`--severity {P0-P3}` (orthogonal to `--priority`; triggers Ed25519 attestation)"*), and `cleo add --severity` **does** persist the value correctly — `cleo show <id> --full` reports `severity: P1` on a task created with it. So the data is there; only the read path drops it.

## Suggested fix

1. **Apply the filter.** It is presumably a missing `WHERE severity = ?` or an unmapped argument between the CLI parser and the query builder.
2. **Validate the value** and exit `E_VALIDATION` on anything outside `P0|P1|P2|P3`. An unrecognised filter value must never widen a result set.
[...]

PR #1330, opened 2026-09-13
fix(T12174): nexus status refuses a foreign path instead of answering about it (gh#1329)

kryptobaseddev:
Addresses the FIRST half of #1329 — **deliberately not `Closes`**.

#1329 names two defects and this PR fixes one:

- ✅ **the reporting bug** — `nexus status` no longer answers about the ambient project when asked about another. It refuses, because a cross-project answer is not available from here.
- ❌ **the parsing bug** — "reject unknown flags on this subcommand so `--output` errors instead of becoming a positional" is untouched. That belongs to the strict-flags chokepoint (#1276), and the issue itself groups it with #1245.

A closing keyword here would have marked #1329 done while the swallowed-flag half — the route by which the reporter *arrived* at the defect — was still live. A wrongly-closed issue is invisible in exactly the way this backlog is already full of, so the reference is non-closing on purpose..

## A confident yes for a project that does not exist

```
$ cleo nexus status /definitely/not/a/real/repo --json
success=true  indexed=true  nodes=26964
repoPath=/definitely/not/a/real/repo
projectId=L2RlZmluaXRlbHkvbm90L2EvcmVhbC9y
```

A path that does not exist reported **`indexed: true`** with the real project's full node count, under a `projectId` derived from the bogus path.

## The core was honest; the CLI was not

`getIndexStats` documents its parameter as **`_projectId`** — deliberately unused since ADR-090/T11648, because the graph DB is project-scoped: one store per project, and `getNexusDb()` opens *this* project's store.

The CLI derived a `projectId` from a user-supplied path, handed it to a function documented as ignoring it, and reported that identity beside counts it did not scope. **The underscore was the honest admission sitting one layer below the surface that lied.**

## Why this isn't a cosmetic mislabel

`CLEO-INJECTION.md` makes this the **mandated first call** for the whole nexus subsystem:

> **FIRST CALL IS `cleo nexus status`.** … a symbol added after the last `analyze` run is simply absent, and a lookup for it returns `E_NOT_FOUND` that reads identically to "no such symbol".

The command exists so an agent doesn't misread `E_NOT_FOUND`. A confident false `yes` here produces exactly the failure the instruction was written to prevent — through the command written to prevent it.

## Verified on a compiled CLI

| invocation | before | after |
|---|---|---|
| `nexus status /not/a/repo` | `indexed: true`, 26,964 nodes | **`E_NEXUS_CROSS_PROJECT_STATUS`** |
| `nexus status --output envelope` | queried `<cwd>/envelope` | **refused** |
| `nexus status` | works | unchanged (`indexed: true`, 26,964) |
| `nexus status <current root>` | works | unchanged |

**Two entrances, one guard.** `--output` is not declared on this subcommand, so citty consumes its *value* positionally — `--output envelope` silently asked about `<cwd>/envelope`. The cross-project guard catches that route too, so the parsing quirk can no longer produce a *wrong answer*. That's why this PR does not also change flag handling: the dangerous half is closed, and the remaining half is cosmetic.

The refusal names the reason **and** the remedy (`cd <path> && cleo nexus status`) rather than only rejecting — a refusal that doesn't say what to do instead is the geometry this repo already paid for in #1278.

## Tests

They pin the two **preconditions**, not the message: that `path` is declared positional (so the hazard exists and a guard must), and that no `output` flag is declared (which is what makes the swallow possible). Pinning the error string would lock in wording; pinning the preconditions means the test still means something if the wording improves.

95/95 across the nexus suites. Typecheck 0.

## Housekeeping

Staged file-by-file rather than `git add -A` — the build dirtied `command-manifest.ts` again (#1306), and that artifact belongs to #1312. Second time today that guard has earned itself.

---

## The three CI failures were gh#1308, not this branch's code (2026-09-12)

    FAIL: Stdout Write Allowlist Lint (T9924)
    FAIL: Stdout Discipline Lint (T10135)
    FAIL: Arch Boundary Check          ← only the aggregate re-reporting the two

Both gates keep **separate** baselines keyed on `path:line`. Any edit that shifts
lines above a pre-existing baselined violation re-reports it as NEW, at a site
the change never touched — here `nexus.ts:1600` and `:1601`, a rendering block
now two lines lower because the cross-project guard sits above it.

Regenerated both.

### Proof the churn is pure re-keying, not an absorbed regression

The counts are unchanged, and that is exactly why a count check is not the proof:
`50 == 50` cannot distinguish "nothing changed" from "one violation removed and
one added". Each entry was instead resolved to the **source statement** it points
at, before (`origin/main`) and after, and the multisets compared:

| gate | total | distinct statements | content-added | content-removed |
|---|---|---|---|---|
| stdout-discipline (T10135) | 50 → 50 | 46 → 46 | **0** | **0** |
| stdout-write-allowlist (T9924) | 55 → 55 | 51 → 51 | **0** | **0** |

Every baselined violation still points at the same statement in the same file;
only line numbers moved. `distinct` < `total` because a few identical statements
recur at several lines in one file — the comparison is a multiset, so
multiplicity is preserved rather than collapsed.

### Why this is worth recording rather than just fixing

The only action that clears this gate's false positive is the same action that
silently absorbs a true one — and the more often it fires, the less scrutiny each
regeneration gets. That is a property of keying on `path:line`, not of anyone's
carefulness: a baseline keyed on statement identity would not have fired here at
all and would still have caught a real addition. Until it is re-keyed, every
regeneration needs a proof like the above attached — a per-incident tax the gate
design imposes (gh#1308).


kryptobaseddev:
Closing as **subsumed by #1341** (`3f9169083`). Not a rejection — every line of this PR is either already on `main` or has been superseded by it, and merging it now would land a failing test.

## Measured, `refs/pull/1330/head` vs `origin/main` at `3f9169083`

```
.changeset/nexus-status-scope-honesty.md       IDENTICAL — already on main
packages/cleo/src/cli/commands/nexus.ts        only in #1330: 2 lines · only in main: 92
.../__tests__/nexus-status-scope.test.ts       only in #1330: 7 lines · only in main: 59
```

**`nexus.ts` — main is a strict superset and this PR's two unique lines are a regression.** They are the raw stdout writes:

```js
process.stdout.write(result.content);
if (!result.content.endsWith('\n')) process.stdout.write('\n');
```

`main` carries the same two statements at `nexus.ts:1690-1691` **with the markers #1341 landed**:

```js
process.stdout.write(result.content); // stdout-discipline-allowed: raw graph payload // stdout-write-allowed: raw graph payload
```

Taking this branch's side reintroduces the unmarked form and re-trips both stdout gates that #1341 fixed by lowering their baselines (50→48, 55→53) so those entries left permanently.

**The test is the dispositive part.** This PR's seven unique lines are one test:

```js
it('declares no `output` flag — so `--output <x>` is swallowed as the path', () => {
  // … the absence of the flag is the precondition and is worth pinning.
  expect(args['output']).toBeUndefined();
});
```

It asserts the **absence** of a flag that #1341 deliberately **added** — main's `status` command declares `output:` directly. So a keep-both resolution lands a red test on `main`.

That test was correct when written. It pinned the *precondition* for the defect: `--output`'s value was consumed positionally because the flag was undeclared. #1341 removed the precondition by declaring it. **The test did not become wrong — its subject was eliminated.** Its own comment says as much: *"the absence of the flag is the precondition and is worth pinning."*

## Why this is a close rather than a merge

The working rule through nine merges tonight has been **"if both sides are real, keep both"** — and it has been right every time, including two cases where a plausible pick would have silently deleted a live subcommand. This is the one place it does not apply, and the distinction is checkable rather than a judgement call: **this branch's content is strictly contained in `main` except where `main` is newer.** Containment can be demonstrated; "both sides are real" is the premise that fails here, because one side was deliberately superseded by the other — same author, same defect, later PR.

## Stated limits

The claim that main's test file passes and this one would not is derived from reading the assertion against main's declaration, **not from executing either**. The evidence is dispositive without execution — the test asserts a key is undefined and main defines it — but it was not run.

Reopen if any hunk here turns out to carry something main lacks; the analysis above is file-by-file and checkable.

kryptobaseddev:
Subsumed by #1341 — see the file-by-file analysis above. No unique content remains.

Issue #1338, opened 2026-09-13
Gate 20 (arch-gate parity) passes on a gate script that is in neither list — absent from both satisfies parity by mutual absence


kryptobaseddev:
## Summary

Gate 20 (`scripts/lint-arch-gate-parity.mjs`, T12122) is the gate on the gates: it asserts that the gate list bundled into `cleo check arch` and the AGENTS.md table name the same set of scripts. It does that correctly.

But a gate script present in **neither** the bundle nor the table satisfies parity **by mutual absence**. So the one state gate 20 cannot detect is the state it most needs to: a gate that exists, is committed, is executable, and that nothing runs.

## Measured

Found while merging #1263, which adds `scripts/lint-cli-startup-barrel.mjs` (T12138 · gh#1207):

```
$ node scripts/lint-arch-gate-parity.mjs
✓ arch-gate parity: 21 gate(s) bundled in `cleo check arch` == 21 documented in AGENTS.md.

$ grep -rn 'lint-cli-startup-barrel\.mjs' --include=*.ts --include=*.mjs --include=*.yml --include=*.md .
.changeset/cli-startup-barrel.md:29:A new gate, `scripts/lint-cli-startup-barrel.mjs`, walks the CLI entrypoint's
```

The only reference anywhere was its own changeset. The script worked — it ran clean and reported `16 modules, zero barrel imports` — it was simply never invoked by anything. Parity reported ✓ throughout, and was right to: 21 == 21. The script was outside both sets being compared.

## Why this is structural rather than an instance

Gate 20 compares two lists to each other. Neither list is `scripts/lint-*.mjs` on disk, so nothing anchors either list to reality. The failure mode is the same family as a parity check between two caches with no read-through to the source: consistent, and consistently able to agree on an omission.

It also has the property that makes this kind of gap persist — **the gate reports success in exactly the case that needs attention.** An unregistered gate produces no output anywhere, so there is no signal at all: not a warning, not a skip, not a count that looks low. It is indistinguishable from a gate that does not exist.

This is the same shape as the `catch { continue; }` in gate 21 and as #1259's memory bound (computed, discarded, everything green): *the right computation exists, is correct, and is not reached by the thing that decides.*

## Suggested fix

Add a third set to the comparison: enumerate `scripts/lint-*.mjs` on disk and assert every one is either registered in the bundle **or** carries an explicit, reasoned exemption.

[...]

PR #1427, opened 2026-09-15
fix(gh#1373): gates 14+15 validate FLAGS, not just verbs

### LINKED ISSUE #1373: Gates 14/15 validate cleo VERBS but never FLAGS — #1276 turns every latent no-op flag into a hard failure
## What

Gates 14 and 15 assert that every `cleo <verb> [<sub>]` named in `CLEO-INJECTION.md` and in workflow `run:` blocks **resolves against the command manifest**. Neither asserts anything about the **flags** passed to those verbs.

citty silently ignores unknown flags, so a flag that never existed has always been a no-op that nothing reports. #1276 turns unknown flags into a hard `E_UNKNOWN_FLAG`. It does not break these call sites — **it reveals them**, and it does so in the release that ships it.

## Measured

Swept every `cleo <verb> … --flag` in `.github/workflows/*.yml` and `scripts/*.mjs` (19 distinct verb+flag groups) and tested the suspects against a build at `2f2a2fc2e`:

```
cleo init --yes                                        -> E_UNKNOWN_FLAG   (rejected)
cleo worktree prune --orphaned --idle-days 7 --yes --json -> rc=0          (valid)
```

One live instance in the repo's own workflows:

```
.github/workflows/worktree-cleanup.yml:73
  cleo init --yes 2>/dev/null || pnpm dlx @cleocode/cleo init --yes 2>/dev/null || true
```

`--yes` does not exist on `init` and never has. On `2026.8.9` it is accepted and ignored (rc=0, `.cleo/cleo.db` created). Post-#1276 both branches of that `||` chain fail, the trailing `|| true` swallows it, and **the step silently does not init** while continuing to report success.

Also found in two verification scripts written tonight (both fixed): the same `cleo init --yes`, where it caused every subsequent command to fail — and the scripts still reported PASS, because they asserted only that the process exited.

`cleo version-bump` still appears in four files but every occurrence is a comment explaining its removal (T12093), which is gate 15 working correctly.

## Why this is a release-timed risk rather than a latent one

Every latent no-op flag in every script, workflow, runbook and consumer project becomes a hard failure **at the moment v2026.9.1 is installed**. The failures appear in the consumer's tooling, not in this repo's CI, and the ones wrapped in `|| true` (like the case above) do not appear at all — they just stop doing their job.

This repo's own workflows are the tractable part; the surface that cannot be swept is downstream.

## The structural gap

The existing gates check the **verb** half of `cleo <verb> --flags` and stop there:

| gate | asserts | does not assert |
|---|---|---|
| 14 (T12069) | verbs in `CLEO-INJECTION.md` resolve | any flag exists |
| 15 (T12093) | verbs in workflow `run:` blocks resolve | any flag exists |

Gate 15 exists because `cleo version-bump` cost a full green preflight per dispatch to discover one missing verb. A missing *flag* had exactly the same invisibility until #1276, with the additional property that it fails **silently** rather than loudly.

The manifest already carries each command's flag list — the `E_UNKNOWN_FLAG` error prints it, which is how the two cases above were confirmed — so extending either gate to validate flags is a matter of joining against data that already exists.

## Not proposing a fix

Whether this belongs as an extension to gate 15, a third gate, or a one-off sweep before the release is a scoping call. The immediate concrete item is `worktree-cleanup.yml:73`.

kryptobaseddev:
Closes #1373 (structural half — the concrete `worktree-cleanup.yml:73` item landed earlier as #1376).

Gates 14 and 15 asserted the **verb** of `cleo <verb> [<sub>] --flags` and stopped there. Both now join each invocation's flags against the command's own `args` declaration.

## Most serious finding: the rollback path has been losing data silently

`packages/core/templates/workflows/release-rollback.yml.tmpl` passed `--reason "${REASON}"` to `cleo release reconcile`, whose entire flag surface is `--from-workflow --rollback --dry-run --json`. mri defaults an undeclared flag to boolean, so `"${REASON}"` fell through as a stray positional.

**The operator's rollback reason has never reached the database** — in the step whose own comment reads *"reconcile MUST write rollback marker into `releases` + `release_changes`"*, which reported success every time, in a template shipped to every consuming project.

This is a silent failure under a success report in the rollback path specifically: the one place you find out during an incident, when you are reading provenance to work out what happened. `reconcile --rollback` has no way to record a reason at all (its own help marks `--rollback` "deferred to T9528"), so the fix drops the flag, echoes the reason into the run log so it is not lost from where an operator can see it, and says plainly that it is not persisted.

## The premise the issue was written on is wrong, and it changes the picture

#1276 (T12139) is described as turning every unknown flag into a hard `E_UNKNOWN_FLAG`. Measured against the **shipped 2026.9.3 binary**:

```
cleo list --zzzbogus            -> E_UNKNOWN_FLAG rc=6
cleo find --zzzbogus            -> E_UNKNOWN_FLAG rc=6
cleo memory find --zzzbogus     -> accepted rc=0
cleo worktree list --zzzbogus   -> accepted rc=0
cleo backup list --zzzbogus     -> accepted rc=0
cleo docs list-types --zzzbogus -> accepted rc=0
```

`assertKnownFlags` has exactly two callsite classes: `lazy-command.ts:87`, which wraps **manifest** entries and passes `meta.name`, and five hand-wired calls in `docs.ts` from T10359. citty dispatches a `verb sub` invocation to the sub-command's own `run`, which the wrapper never reaches.

**The guard covers root verbs only.** Across the entire `verb sub` surface the silent-no-op defect is still fully live today. The gate reports the two differently rather than blurring them:

| `guard` | meaning |
|---|---|
| `rejects` | hard failure the moment the caller upgrades |
| `silent` | accepted and discarded — the `cleo list --severity P0` shape: a confident unfiltered answer indistinguishable from a real one |

## Defects found

**Shipped templates** (no deployed counterpart — consumers only):

- `release-rollback.yml.tmpl` passed `--reason "${REASON}"` to `release reconcile`, whose whole flag surface is `--from-workflow --rollback --dry-run --json`. mri defaulted the undeclared flag to boolean and `"${REASON}"` fell through as a stray positional. **The operator's rollback reason has never reached the database**, while the step whose own comment says it "MUST write rollback marker" reported success every time.
- `release-publish.yml.tmpl` told an auto-filed incident issue to run `cleo release reconcile <v> --backfill`. No such flag — and the same file already had the correct spelling 50 lines earlier.

**Emitted by code, not merely documented:**

- `injection.ts:200` wrote `# Run: cleo memory digest --brief` into the AGENTS.md of every project in cli bridge mode. `digest` has no `--brief`; it produced a digest anyway because that *is* the default, which is exactly why nothing ever reported it.
- `suggested-next.ts:123` returned `cleo orchestrate ready --epic <id>` as CLEO's own next-step advice. `ready` takes EPICID as a **positional**; it worked by accident of mri's boolean defaulting.

**CLEO-INJECTION.md**, injected verbatim into every spawned agent:

- `orchestrate start --epic`, `orchestrate ready --epic` (twice) — both positional commands.
- `memory decision-find --epic` — promised an epic filter that does not exist; the flag was accepted and the **unfiltered** decision set returned.

`suggested-next-auto-populate.test.ts` pinned the emitter's old output. The code was what was wrong, so the expectation moved with the fix.

## Verification

- **Extractor validated against the shipped binary on 71 leaf verbs** by passing `<claimed-valid-flag> --zzzbogus`, which exits at the flag check without running the command: **41 OK, 0 MISMATCH**. The 16 inconclusive fail on a missing required positional, which citty validates before `run()` is reached.
- Every finding independently confirmed against the binary's own `--help` flag list — a different *kind* of question from parsing source.
- **Fail-injection both directions**: three known-bad flags fail with correct per-command attribution; a `#` comment naming a removed flag does not.
- Gate 20 parity untouched (22 == 22). These extend existing gates, so there is no new registration for #1338's mutual-absence hole to swallow.

Two bugs in this gate were caught before it gated anything, and both have tests: the global flag list parsed as **empty** (anchored on the `[]` of the type annotation rather than `Object.freeze([`), which would have reported `--json` as unknown on every command; and a markdown tail that ran past a backtick, merging two code spans so one command's flag was blamed on another — the gate still fired, at the wrong command, with one defect hidden.

Not expanded into here, measured and filed separately: **70** unknown-flag hits across `packages/skills/skills/**` (including tier-0 `ct-orchestrator` and `ct-cleo`) and **49** across `packages/*/src/**/*.ts`.

Tests were not run locally — an unbounded vitest fork pool has frozen this machine twice. All 33 assertions were verified through a plain-node harness against the real repo sources; CI runs the vitest file.


kryptobaseddev:
Landed via the integration batch #1434 (merge `a1a232da3`) rather than on its own. Verified by content, not by the merge label:

```
scripts/lint-injection-commands.mjs                          main=dbdc2fc94c == batch=dbdc2fc94c
scripts/__tests__/lint-flag-existence.test.mjs               main=2c65b61427 == batch=2c65b61427
packages/core/templates/workflows/release-rollback.yml.tmpl  main=5f11e2a998 == batch=5f11e2a998
packages/core/src/dispatch/suggested-next.ts                 main=7cb39eded0 == batch=7cb39eded0
```

One commit exists in the batch that is not on this branch — `aba0bfd139`, also mine, and worth recording why:

This branch's template test asserted `toEqual(['/data/task/verification', '/data/task'])`, a **snapshot** of the template as it stood here. #1430 corrected `CLEO-INJECTION.md:279` and removed the `never --field /data/task` negative example — both correct edits — and my assertion failed on them. The test was wrong, not the edit. It now asserts the property it was always named for: verification pointers are attributed to `verify`, never to the earlier `cleo show` on that line.

Neither PR could have caught it alone: `Scripts Tests` is path-filtered on `scripts/**`, #1430 touches none of that path, so the job **skipped** there. My tests and their template only ever met inside the batch.

The general rule, since `CLEO-INJECTION.md` is edited by whoever is fixing the protocol and rarely by the person who wrote the test: a snapshot of its contents is a tripwire under someone else's correct work.

T12191 recorded against `pr:1434` and completed.

Issue #1439, opened 2026-09-15
Gate 19 (CLI startup barrel imports) has been BREACHED — 109 vs baseline 106 — and no workflow runs it, so it could not have gone red


kryptobaseddev:
## The breach

```
$ node scripts/lint-cli-startup-barrel-imports.mjs
lint-cli-startup-barrel-imports: FAIL — 109 static core-barrel import(s), baseline 106.

New since baseline:
  • packages/cleo/src/cli/commands/config/show.ts:16      imports '@cleocode/core'
  • packages/cleo/src/cli/commands/config/validate.ts:16  imports '@cleocode/core'
  • packages/cleo/src/backfill/audit-columns.ts:29        imports '@cleocode/core/internal'
```

Gate 19 is a **ratchet**: the count of static `@cleocode/core` barrel imports in the CLI may fall but never rise. It has risen by three. Each one forces the full 1266-module core graph to load before any command runs — measured at 2.54 s for the barrel vs 0.12 s for a deep subpath module.

Found by cleo-dev running every `scripts/lint-*.mjs` against an integration tree; reproduced here independently against `origin/main`.

## Why nobody could have known

| surface | gate 19 present? |
|---|---|
| AGENTS.md arch-gate table (row 19) | **yes** |
| `cleo check arch` bundle (`check.ts:569`) | **yes** |
| any `.github/workflows/*.yml` | **no** |
| any template workflow | **no** |
| root `package.json` scripts | **no** |

```
$ grep -rn "lint-cli-startup-barrel-imports" .github/workflows/ packages/core/templates/workflows/
(no matches)
```

Note the subtlety that makes this invisible to the gate-on-the-gates: **gate 20 (arch-gate parity) passes**, correctly. It joins the `cleo check arch` bundle against the AGENTS.md table, and gate 19 is in both — 22 == 22. Parity is satisfied by a gate that is documented, bundled, and executed by nothing.

And the CI job that sounds like it would run it does not. `Arch Boundary Check` invokes **individual scripts** (`node scripts/lint-no-direct-db-open.mjs --strict`, `node scripts/lint-publish-surface.mjs`, …) rather than `cleo check arch`. Gate 19's script is not among them.

So the only thing that ever executes gate 19 is a human or agent typing `cleo check arch` locally — which AGENTS.md does instruct, but which nothing enforces and which produces a failure nobody is obliged to act on.

[...]

Issue #1381, opened 2026-09-14
A tool killed by a signal reports "binary missing or spawn error" — the memory guard OOM-kills `tool:test`, and the error sends you hunting a binary that is present


kryptobaseddev:
## The observation, from two independent lanes on another repo

Both running `cleo verify <id> --gate testsPassed --evidence "tool:test"` on `v2026.9.1`:

```json
{"success":false,"error":{"code":1,
  "codeName":"E_EVIDENCE_TOOL_UNAVAILABLE",
  "message":"Tool \"test\" → npm test could not be executed (binary missing or spawn error)"},
 "meta":{"duration_ms":2456148}}      // 40.9 minutes
```

One lane traced the process tree from `/proc` at the 12-minute mark and again at 22 minutes:

```
cleo verify (475576)
  └─ npm test (478853)                                              ← wrapper
       └─ node .../pnpm -r --no-bail --if-present run test (478887) ← the real command
            └─ seven vitest workers, executing
```

**The command resolved, spawned, and ran for forty-one minutes.** Nothing was missing.

## What is actually happening

Not the hypothesis the lanes reached (that a non-zero exit from `pnpm -r --no-bail` maps onto `E_EVIDENCE_TOOL_UNAVAILABLE`). That branch is correct in the source — `evidence.ts` classifies `exitCode !== 0` as `E_EVIDENCE_TOOL_FAILED` and attaches a 512-byte tail. The reports are right about the symptom and wrong about the cause, which matters because the real cause is ours and the hypothesised one is not.

The chain:

1. **`tool-cache.ts:838-849`** — `test` and `build` are wrapped by `withMemoryLimit()`, which spawns them inside a transient systemd scope with a hard `MemoryMax` and `MemorySwapMax=0` (T12116). The cap is 25% of RAM, floor 2 GB, ceiling 8 GB (`heavy-tool-limit.ts:55-61`). **Denying swap is deliberate** — the failure it guards against was a throttle-and-thrash host freeze, not an OOM kill.

2. A large monorepo suite — here, seven concurrent vitest workers — exceeds that ceiling. With swap denied, the kernel OOM-kills the cgroup. **The guard is working as designed.**

3. **`tool-cache.ts:558-560`** — the child dies by signal, so Node's `close` event fires with `code = null`:
   ```ts
   child.on('close', (code) => {
     finalise(code);
   });
   ```
[...]

PR #1461, opened 2026-09-16
fix(gh#1396, gh#1397): do not double-wrap systemd-run; name the scope; a harness that never started is not a red suite

### LINKED ISSUE #1396: [Bug]:tool:test is unrunnable: systemd-run transient scope name collides, 4 of 4, and the suite never starts
### Bug Report

ENVIRONMENT: axiom-analytics, CLEO root /mnt/projects/axiom-analytics, git repo in axiom-app/ subdirectory. Command pinned in .cleo/project-context.json testing.command:
  systemd-run --user --scope --quiet -p MemoryMax=8G -p MemorySwapMax=0 -- env VITEST_MAX_WORKERS=4 NODE_OPTIONS=--max-old-space-size=2048 pnpm --dir axiom-app exec vitest run --maxWorkers=4

SYMPTOM: four consecutive runs of
  cleo verify T167 --gate testsPassed --evidence tool:test
returned E_EVIDENCE_TOOL_FAILED, exit 1, with tail:
  Failed to start transient scope unit: Unit run-p<PID>-i<n>.scope was already loaded or has a fragment file
Observed unit names: run-p4134869-i25107050, run-p111620-i25168473, run-p147714-i25169371, run-p2928203-i28028029.

THE SUITE NEVER STARTS. No vitest process is ever spawned. One attempt failed in 11 seconds with the per-tool semaphore slot free, so it is deterministic and unrelated to queueing or contention.

THE COMMAND ITSELF IS FINE. Run verbatim through sh -c it exits 0 and the suite runs. Eight concurrent transient scopes and ten rapid sequential ones collide zero times. systemctl --user reset-failed does not help. A trivial systemd-run --user --scope succeeds at the same moment cleo fails.

CORRELATION, EXPLICITLY UNPROVEN: this host has wrapped its PID space. pid_max 4194304, highest running PID 4176594 (99.6 percent), a freshly spawned PID already back to 129453. systemd-run derives the scope unit name from the caller PID. Attempt 1 took a name near the ceiling, attempt 2 a low post-wrap one. I cannot explain why the -i<n> suffix fails to disambiguate, so this is a hypothesis and not a mechanism.

IMPACT: testsPassed is unsettable for every task in the project while this holds. 21 tasks here have testsPassed as their only outstanding gate.

ASK: when the tool wrapper fails to START, surface that as a distinct error rather than as a tool failure - see the companion issue about E_EVIDENCE_TOOL_FAILED being indistinguishable from a red suite. Consider naming the scope explicitly (--unit=) rather than letting systemd derive it from a PID that can recycle.

**Severity**: P2
**Area**: evidence/tool-resolution

---

## Environment
| Component | Version |
|-----------|---------|
| CLEO | 2026.9.1 |
| Node.js | v24.16.0 |
| Shell | /bin/bash |

### LINKED ISSUE #1397: [Bug]:E_EVIDENCE_TOOL_FAILED cannot distinguish a harness that never started from a genuinely red suite
### Bug Report

Both of these produce E_EVIDENCE_TOOL_FAILED with exit code 1 and nothing else to tell them apart:
  A. the test suite ran and tests failed
  B. the tool wrapper could not start at all, so zero tests ran

On 2026-09-13 I hit B four times (systemd-run could not create its transient scope; filed separately) and A once, in the same hour, on the same project. The envelopes are the same shape. The only way I could tell them apart was to read the raw stdout tail by eye and notice that one said 'Failed to start transient scope unit' while the other said 'Test Files 2 failed'.

WHY THIS MATTERS MORE THAN IT LOOKS: a gate that cannot be set because the harness could not start looks exactly like a gate that cannot be set because the code is broken. An agent that trusts the exit code will go hunting a test failure that does not exist. On this day BOTH were true at once - the harness was broken AND main was genuinely red on two files - and neither was legible from the result.

A RELATED CASE FROM THE SAME DAY, same root shape: pnpm typecheck exited 139 (SIGSEGV) while printing ZERO diagnostics. Any check that greps stdout for 'error TS' reads that as clean. It is not clean; it is a process that died before it could disagree. Absence of diagnostics is not absence of errors when the exit code says the process crashed.

ASK, in order of value:
 1. A distinct code for 'the tool could not be started' vs 'the tool ran and reported failure'. E_EVIDENCE_TOOL_UNSTARTABLE or similar.
 2. Record the terminating SIGNAL alongside the exit code. A child killed by a signal reports through a different field, and a wrapper that binds only code discards it. 139 and 137 are signals wearing exit codes.
 3. Where the tool emits a machine-readable result (vitest --reporter=json), prefer parsing it over the exit code, and say which was used.

VERIFICATION NOTE FOR WHOEVER TAKES THIS: count FILES, not just tests. A test file that throws at module scope fails the FILE and produces ZERO failed assertions, so a test-only count understates a whole-file breakage. My red run reported '1 failed test' for a two-file failure.

**Severity**: P2
**Area**: evidence/error-taxonomy

---

## Environment
| Component | Version |
|-----------|---------|
| CLEO | 2026.9.1 |
| Node.js | v24.16.0 |
| Shell | /bin/bash |

kryptobaseddev:
Closes #1396. Closes #1397. Task: T12221.

## The defect is the double wrap, not the name collision

CLEO wraps `test` and `build` in a `systemd-run` scope (T12116). It never checked whether the command it was handed already had one. `axiom-analytics` pins:

```
testing.command = "systemd-run --user --scope --quiet -p MemoryMax=8G -p MemorySwapMax=0 -- env … pnpm exec vitest run"
```

`parseCommandString` splits on whitespace, so `command.cmd === 'systemd-run'` and CLEO prepended a second one. `systemd-run --scope` **execs** its payload rather than forking, so the inner client regenerates a unit name the outer has already registered.

Measured on systemd 259 (259.8-1.fc44):

| | Result |
|---|---|
| `systemd-run … -- systemd-run … -- /bin/true` (nested) | **5/5 fail**, `Unit run-p<pid>-i<id>.scope was already loaded or has a fragment file` |
| `systemd-run … -- /bin/true` (single) | **0/6 fail** |

That explains the deterministic 4-of-4 the reporter saw, why an idle box reproduces it (it is not contention), and why this repo never could — cleocode's own test command is `pnpm run test`, so nothing double-wraps here.

Hypotheses eliminated on the way, recorded so nobody re-walks them: PID reuse (`pid_max` 4194304 at ~125 PID/s → ~9.2 h per wraparound, far too slow for a deterministic failure), lingering uncollected scopes (they self-reap), and env sanitisation (`spawnCmd` spreads `process.env` first and whole).

**So the primary fix is detection, not naming.** Declining to re-wrap is the rule this codebase already applies twice: `heapCapApplied()` does not re-exec over an operator's own `NODE_OPTIONS` heap cap, and `mergeNodeOptions()` lets an existing explicit value outrank our default. A project pinning `-p MemoryMax=8G` has made exactly that choice, and a second ceiling it did not ask for is the thing to avoid.

Detection is **systemd-run-specific**. `env`, `nice`, `taskset` and `timeout` are plausible leading tokens too, but none imposes the cgroup memory bound this function exists to apply, so treating them as already-confined would silently drop the ceiling. They are named as known-unhandled in the TSDoc.

## `--scope` and `--unit=` are not alternatives

The scope CLEO *does* create is now named `cleo-tool-<canonical>-<rootHash8>-<rand8>.scope`.

`--scope` selects the unit **type**; `--unit=` supplies its **name**. They are independent and compose — `spawn-wrapper.ts` has always passed both. The TSDoc at `heavy-tool-limit.ts:167` asserted they were alternatives, and that one sentence is the whole defect.

| Component | Defends against |
|---|---|
| `cleo-tool-` | *(legibility)* `systemctl --user list-units 'cleo-tool-*'` enumerates exactly CLEO's heavy-tool scopes |
| `<canonical>` | *(legibility)* which tool is stuck, without opening anything |
| `<rootHash8>` | two worktrees of one repo; two projects on one host |
| `<rand8>` | concurrent verifies of the same tool in the same tree; PID reuse |

The PID is **deliberately absent** — it is the component that makes systemd's auto-generated name collide and defends nothing randomness does not. `rootHash8` comes off the same execution root the cache key uses (T12112 / gh#1220), so the scope and the cache agree on what "the tree under test" means.

## A second defect closes with it: heavy-tool scopes were unreapable

`suite-reaper.ts:93` reaps via `systemctl --user stop <unitName>`. It can only stop units it can **name**. Every heavy-tool scope was auto-named, so the reaper that already exists could never target one. Naming makes them enumerable and individually stoppable — there is a test that starts a scope and stops it with exactly the command the reaper runs.

## gh#1397 — the silencer, fixed first on purpose

`systemd-run --scope` is transparent on success: it exits with the wrapped command's status (a scope running `sh -c 'exit 7'` makes `systemd-run` exit 7). When it cannot create the unit it exits **1**, and the payload never runs. `1` is also what a suite with a failing test exits with.

So "the harness never started" and "the suite ran and was red" were identical in everything `validateTool` examined, and every harness failure was reported as `E_EVIDENCE_TOOL_FAILED` — a suite that failed in about two seconds. Five occurrences of #1396 produced no diagnosis for exactly that reason.

`E_EVIDENCE_TOOL_UNAVAILABLE` was **always** the right code. What was missing was a route into it: the existing guard keys on `exitCode === null`, and `systemd-run` itself starts perfectly well before failing. Harness failures now route there, quote systemd's own diagnostic, and are **not cached** — the cache key cannot rotate on an unchanged tree, so a persisted harness failure would serve a fabricated "your tests failed" forever without ever spawning again.

`spawnCmd` also stops discarding Node's spawn-error object — the same defect #1381 fixed one event-handler over, for `signal`.

## The near-regression: why the gate asks a wider question than it looks like it should

**This is the part worth reading.**

The harness-failure classifier was first gated on `limited.confined` — "did CLEO wrap this?". That is the obvious condition, and it is wrong.

Detection makes `confined` **false** for exactly the `axiom-analytics` shape. So the #1396 fix would have silently reintroduced #1397 **for the one project that reported it** — a fix composing with an adjacent fix to restore the original symptom in a narrower, harder-to-find form. No diff review finds that; you only find it by asking what the gate's condition *means* after the change.

The gate now asks whether the spawned command was a `systemd-run` invocation **at all**, whoever added it: `limited.confined || isSystemdRunCommand(command.cmd)`. The TSDoc records why the narrower condition is wrong, because the narrower condition is the one someone will propose again.

## Tests

18 new tests, all green.

Proof they fail on unfixed code — behaviour disabled surgically (value nulled, `evidence.ts` arm removed) with types and imports intact, so **assertions** fail rather than imports:

| | Result |
|---|---|
| Fixed | **9 / 9 pass** |
| Unfixed | **3 fail / 6 pass** |

The three that fail are the fix-detecting assertions. The six that still pass are behaviour-preservation controls — including *a genuinely red suite stays `E_EVIDENCE_TOOL_FAILED`*. A test that went red on everything would only prove something was broken.

The two `validateAtom` cases are a deliberate **pair**: both run a tool that exits `1`, and only stderr differs. A fix keying on exit code, duration, or absence of stdout would pass one and fail the other.

Two integration cases need a real user systemd manager and use `it.skipIf`, with a `console.warn` naming the reason — a silent skip is indistinguishable from a pass, which is this PR's own subject.

## Scope of the claim

The `--unit=` change is a **necessary condition proven, not a sufficiency proven**. The nesting mechanism is reproduced here 5/5 and the reporter's `project-context.json` is confirmed on disk, but the original 4-of-4 occurrences were observed on another host and are not re-run here.

## Gate status

- **`cleo check arch`: 21 passed, 1 failed.** The failure is `scripts/lint-cli-startup-barrel-imports.mjs` (runner `gate-10`, AGENTS.md row 19 — the two numbering schemes collide, join on script path), at 109 vs baseline 106. **Pre-existing on main, not from this PR**: all three flagged files are in `packages/cleo/`, all are unmodified, and a stash control with every change removed still fails 109/106. Filed as #1439 / T12203. Baseline deliberately **not** raised.
- Typecheck: **0 errors**, none in the touched files.
- biome: clean.
- `scripts` project: baseline **4 failures** measured alone on a clean tree, unchanged by this PR. The four stdout/cwd/changeset gates each exit **0** run directly against the tree.


kryptobaseddev:
Review — one blocking item, and the lint that flags it is giving advice that would silently break this PR.

## CI failure is real and is from this PR

`Project Root Anti-Pattern Lint (T9584)` fails. Reproduced locally, exit 1:

```
[core-process-cwd] packages/core/src/tasks/heavy-tool-limit.ts:317
  const unitName = buildToolScopeUnitName(canonical, opts.executionRoot ?? process.cwd());
  -> Bare `process.cwd()` in packages/core/src bypasses `getProjectRoot()` —
     use `resolveOrCwd(...)` (T9584) or annotate with `// CWD-OK: <reason>`
```

66 pass / 1 fail / 11 pending at time of writing. This one is introduced here, not pre-existing — the line is new in this PR.

## Do NOT take the fix the lint suggests

The lint recommends `resolveOrCwd(opts.executionRoot)`. That differs from the current code only in the fallback: `getProjectRoot()` instead of `process.cwd()`. Applying it would defeat the purpose of `rootHash8`.

`getProjectRoot()` deliberately collapses a worktree to the main repo. From `packages/core/src/paths.ts`, step 2.5:

> if `start` is inside a git worktree (i.e. has `.git` as a gitlink FILE pointing to `<mainrepo>/.git/worktrees/<name>`), the canonical project root is the **MAIN repo**, not the worktree dir.

So under `resolveOrCwd`, every worktree of a repo would hash to the same `rootHash8`. That is precisely the case the component was introduced to separate — the justification for `rootHash8` is that two worktrees of one repo must get distinct scope namespaces so either can be reaped without touching the other. The lint would reintroduce a collision class in the fix for a collision bug.

This is the same shape as the near-regression already documented in this PR body: a correct-looking local change that re-breaks the thing the change exists to fix.

## Recommended resolution

Annotate, and put the reason at the line where someone will later try to "fix" it:

```ts
const unitName = buildToolScopeUnitName(canonical, opts.executionRoot ?? process.cwd()); // CWD-OK: the scope name must distinguish worktrees; getProjectRoot() collapses a linked worktree to the main repo (paths.ts step 2.5), so resolveOrCwd() would give every worktree of a repo the same rootHash8 and defeat the per-tree namespacing this name exists for. The production call site (tool-cache.ts:1128) always passes executionRoot; this fallback is probe/test-only.
```

Considered and rejected: making `executionRoot` required on the opts. It is the cleaner type, and it would remove the `process.cwd()` entirely, but there is an internal caller at `heavy-tool-limit.ts:379` (`withMemoryLimit(canonical, "probe", [], { env })`) plus a number of test callers that do not pass it. That is a larger change than this PR should carry. Worth a follow-up task, not a blocker.

## One thing to check while you are in there

`heavy-tool-limit.ts:379` builds a limited command for the confinement probe without an `executionRoot`. If that path can reach the naming branch, the probe would construct a unit name from the process cwd. Confirm whether it short-circuits before `buildToolScopeUnitName` — if it does, say so in the annotation; if not, it needs its own answer.

## Everything else reviewed

- Double-wrap detection being systemd-run-specific, with `env`/`nice`/`taskset`/`timeout` named as deliberately unhandled because none imposes a cgroup bound — correct, and the right reason.
- The wider gate question (`limited.confined || isSystemdRunCommand(command.cmd)`) with the rationale in TSDoc — this is the most important change in the PR and it is documented where it will be read.
- Fails-on-unfixed evidence (3 fail / 6 pass, fix-detecting red and behaviour-preservation green) is the correct shape.
- Arch reported honestly as 21/22 with the failure named by script path and both gate numbers, baseline not raised — correct; that failure is #1439 / T12203 and is not this PR to fix.

kryptobaseddev:
Prior art for the `// CWD-OK:` annotation at `heavy-tool-limit.ts:317` — it is the second instance of an established pattern in this file's own neighbourhood, not a one-off exception.

`resolveEvidenceExecutionRoot()` in `packages/core/src/tasks/evidence.ts` (origin/main, line 1370) carries a `// CWD-OK:` comment with the same reasoning, arrived at independently: that `resolveOrCwd()` falls back to `getProjectRoot()`, which deliberately collapses a linked worktree to the main repo, and that using it there would silently revert gh#1220 while every test stayed green.

That is exactly the argument in the review comment above for why `resolveOrCwd(opts.executionRoot)` must not be applied here — a worktree-distinguishing identity cannot be sourced from a resolver that is documented to collapse worktrees.

So a reviewer seeing this annotation should read it as the codebase being consistent with itself: where a value must distinguish worktrees, `getProjectRoot()` is the wrong source and the exemption is annotated rather than silently taken. Two independent authors reached the same conclusion about two different call sites.

Found by cleo-helper while root-causing gh#1365 in the same file.

PR #1481, opened 2026-09-18
fix(ci): repair Core Tarball Size Gate — the pnpm ellipsis was inverted

kryptobaseddev:
## The gate has never passed

`gh run list --workflow=core-tarball-size.yml --limit 100` returns `{"failure": 47}` — **47 runs, 47 failures, 0 successes**, back to v2026.6.10. It dies in `Build @cleocode/core (+ deps)` on five `TS2307`s in `packages/brain`, so it has **never reached its own size assertion**. Nobody knows whether `@cleocode/core` is under its 30 MB budget.

Found while investigating why v2026.9.6/9.7 published green over a package that did not exist — this is the second of three red signals in the release pipeline that nothing reads.

## Cause: the ellipsis direction

```
'@cleocode/core...'  (trailing) = core and everything core DEPENDS ON
'...@cleocode/core'  (leading)  = core and everything that DEPENDS ON core
```

T11976/DHQ-079 switched both filters to the **leading** form and wrote a comment asserting the opposite meaning. The comment is why this survived 47 runs: the filter looks wrong to anyone who reads the prose and right to anyone who trusts it.

Measured read-only in this workspace:

| filter | projects | includes |
|---|---:|---|
| `'@cleocode/core...'` | 17 | paths, contracts, lafs, agents, git-shim, nexus, skills, utils … |
| `'...@cleocode/core'` | 12 | **none of those**; cleo, studio, brain |
| `'@cleocode/core...' + '!@cleocode/studio'` | **16** | all deps, no studio |

The eight packages the leading form omits are exactly what `brain` reported as unresolvable. CI confirms the selection: `Scope: 12 of 22 workspace projects`.

## Why `!@cleocode/studio` stays

`packages/core/package.json:513` declares `@cleocode/studio` in `optionalDependencies`, so `core → studio → brain → core` is a cycle — pnpm warns about it verbatim in the failing log. Studio is therefore pulled into the trailing-form selection too, and excluding it keeps vite/`@sveltejs/kit` out of the build. That is exactly what T11976 was correctly trying to achieve; it just repaired the symptom by breaking the selection.

The cycle is also why `brain` was scheduled **first**, before `paths`/`contracts`/`core` had a `dist/` — which is why all five errors are "module or its type declarations not found" rather than type errors.

## This lands alone, deliberately

Because the assertion has never executed, **the first green run may fail on the 30 MB budget.** Core is 42.8 MB across 5319 files unpacked; its packed size is unmeasured. Adjusting `MAX_CORE_TARBALL_MB` and refreshing the stale v2026.6.14 composition tables is a separate PR, made after reading a number this gate has never printed.

Do not assume this goes green.


kryptobaseddev:
## It passes — for the first time in 48 runs

`workflow_dispatch` on this branch: [run 35373736283](https://github.com/kryptobaseddev/cleo/actions/runs/35373736283) — **success**.

The number this gate has never once printed:

```
@cleocode/core packed: 11.97 MB / budget 30 MB (unpacked 49.21 MB, 5322 files)
     5.33 MB  binaries/fallback/cleo-supervisor
     3.02 MB  binaries/fallback/worktree-napi.linux-x64-gnu.node
     2.41 MB  dist/llm/plugin-facade.js
     2.40 MB  dist/selfimprove/fix-gen.js
     1.48 MB  dist/docs/export-document.js
     1.27 MB  dist/llm/plugin-facade.js.map
     1.26 MB  dist/selfimprove/fix-gen.js.map
OK — within the 30 MB budget; fallback binary policy satisfied.
```

**So no threshold PR is needed after all.** I flagged one as likely because the assertion had never executed and I would not assume green — it is green, with 18 MB of headroom. The budget was never in danger; the gate was just broken.

That is worth stating plainly: for 48 runs across four months this repo had a size gate that could not have caught a regression, and nobody could tell, because a red tag-triggered workflow that gates nothing looks exactly like a red tag-triggered workflow that gates nothing.

Two observations from the output, neither in scope here:

- `plugin-facade.js.map` (1.27 MB) and `fix-gen.js.map` (1.26 MB) are in the top five contributors — core ships sourcemaps, and the workflow's own composition table already names maps as "the first trim lever". That is an install-time optimization, not a publish-delay one.
- The composition table in the header (`v2026.6.14 baseline`) estimated 18–22 MB packed. Actual is 11.97 MB. Stale, but in the safe direction.

PR #1499, opened 2026-09-20
fix(hierarchy): reachable subtask tier, cleo decompose, --auto-decompose, reconcile scope, and a test fork budget that works

kryptobaseddev:
## Why

Two field reports, two different repos, same wall:

```
cleo add --type subtask --parent T473 ...
→ E_CLEO_DEPTH_EXCEEDED: "Use --parent T111 (the parent epic) instead."
```

One agent burned three `cleo add` calls, abandoned the decomposition it had correctly designed, and fell back to `orchestrate spawn`. **`cleo add --type subtask` was structurally impossible in every saga-rooted project.**

Owner also reported two adjacent, long-standing problems: overlapping task scope across sagas/epics accumulating with nothing to reconcile it, and the local test suite repeatedly OOM-killing developer machines.

Tracked under epic **T12295** (saga T12243).

---

## 1 · The depth cap — a stale constant (T12296)

`maxDepth: 3` was calibrated when the spine was `epic(0) → task(1) → subtask(2)`. The config schema still said so verbatim. ADR-083 §2.5 / ADR-088 then made `saga` a real `parent_id` container, pushing every tier down one. The write-path guard was never recalibrated:

```ts
if (parentDepth + 1 >= policy.maxDepth)   // task now at depth 2 → 3 >= 3 → throw
```

**Evidence it was the guard, not the data:** this repo's own store holds **163 live `saga → epic → task → subtask` rows** `cleo add` could no longer create. Four guards had drifted into three calibrations; the two implementing the *documented* model were the two not on the write path.

Fixed by standardising on the inclusive reading and collapsing four copies into one `exceedsMaxDepth`.

**The error also mis-routed callers.** It suggested `--parent <epic>` while the caller's `--type subtask` travelled with the retry, and the matrix only admits `task→subtask` — following the hint moved you from exit 11 to exit 6. Suggestions are now cross-checked against `PARENT_TYPE_MATRIX`.

## 2 · `cleo decompose` + `--auto-decompose` (T12298)

The missing verb behind PM-Core V2 design-point 3. Its documented escape hatch never worked: `cleo update <id> --acceptance ""` is rejected by acceptance enforcement whenever mode is `block`, the default.

`--auto-decompose` collapses the three-command stutter (add → decompose → add) into one. Opt-in, because it rewrites the parent; the envelope reports `autoDecomposed: { childId, movedAcceptance }`.

Also fixes `forceDuplicate` not covering the exact-title path — `addTask` returned an existing task without inserting and without throwing, so a caller who said "create it anyway" silently got back a row it did not create.

## 3 · `cleo reconcile scope` (T12299)

Overlap detection was insert-time, flat, and binary. This adds a retro-active, containment-aware sweep classifying each pair into **merge / absorb / split / link**. Read-only by default; `--apply` writes only `relates` edges — nothing merged, retitled, reparented or deleted.

**Threshold calibrated, not chosen.** Swept saga T9977 (144 live tasks, 10,160 pairs):

| threshold | findings |
|---|---|
| 0.55 | 427 |
| 0.60 | 128 |
| **0.65** | **32** |
| 0.70 | 7 |

427 isn't a report anyone acts on. Residual false positives are documented: `Route createAgentWorktree…` vs `Route destroyAgentWorktree…` scores 0.852 and is two different jobs — which is why merge/absorb also require a tier+parent match. On the real sweep those fired **zero** times while firing correctly on a synthetic control with true duplicates.

## 4 · The test fork budget (T12297)

`MEMORY_SAFE_MAX_WORKERS` divided `totalmem()` by the per-fork budget. Installed RAM says nothing about what's free, so on a shared box the same bytes were budgeted twice and runs were OOM-killed.

Third revision of this guard. T11839 put bounds in the root config where `--root packages/<pkg>` never inherited them; T12087 moved them into a spread module but kept the `totalmem` denominator. Now takes `min(cgroup v2 budget, MemAvailable)` with 12 GB reserved headroom and a `CLEO_TEST_MAX_WORKERS` override.

The cgroup read **walks `/proc/self/cgroup` ancestors** — the limit sits on `app.slice`, and the root cgroup is unreadable, which is exactly where the previous code looked. Verified the walk finds one ancestor with a limit and skips the unreadable root.

---

## Verified end-to-end

| | before | after |
|---|---|---|
| `saga → epic → task → subtask` | **blocked** | builds |
| child under a subtask (depth 4) | blocked | still blocked |
| the error's own `fix` string | → exit 6 | **executes cleanly** |
| `--type subtask --parent <epic>` | — | rejected, not silently retyped |
| subtask under a text-AC task | 3 commands | **1** with `--auto-decompose` |
| overlap across a saga | invisible | 32 classified findings |

`cleo check arch`: **21/22**. The one failure (gate-10, CLI barrel imports) fails identically on `origin/main` from three files this PR never touches.

## Why no test caught the original bug

Every depth test asserted arithmetic over untyped `makeTask` fixtures, so inserting a tier above `epic` could not fail them. New tests are written in **tier names** and fail-injected — the spine test goes red on the old `>=`.

## Deliberately not included

Narrowing `ROOT_TASK_TYPES` to saga-only was attempted and **reverted** — `scaffold-validate.ts` consumes the returned violations and fails the scaffold, silently breaking epic-rooted scaffolding across 26 tests. Filed as **T12301**. Native SQLite/vec0 allocations still escape the per-fork heap cap; filed as **T12300**.


kryptobaseddev:
## 🛑 Handoff — do not merge until the exit-143 blocker is root-caused

Handing this PR to another agent. All work is **committed and pushed** (`a741ce9fb`), working tree clean, CI **83/85**.

Full handoff: `cleo docs fetch pr1499-hierarchy-depth-handoff` (attached to epic **T12295**). Blocker tracked as **T12307**.

### The blocker

`Unit Tests (ubuntu-latest, shard 1)` has failed **5 times** with `exit code 143` + *"The runner has received a shutdown signal"*. 143 is SIGTERM. **The mechanism is not fully explained.** Shard 2 passes every time.

**Do not dismiss it as flaky infrastructure — it already hid three real failures once.** Shard 1 reported 143 while three genuine assertions were failing (injection line cap, spawn-prompt size, schema snapshot). A failing assertion exits **1**, so the 143 was a separate event, and the real failures were only found by grepping the shard log for `×`.

Those three are now fixed in `a741ce9fb` and verified locally (21/21, 25/25, 16/16). The 143 is not.

### Established, so nobody re-derives it

- **Shard membership is NOT the cause.** Computed vitest's real algorithm (`sha1(path relative to config.root)`, sorted, sliced by `calculateShardRange`): main 1621 files → 811 in shard 1; branch 1624 → 812. Exactly **one** file moved in (`hierarchy-depth-spine.test.ts`, 19 tests, 6 ms), **zero** moved out. My own reshuffle hypothesis is falsified.
- **Another branch is not evidence.** `fix/gh1462-evidence-git-root` also failed shard 1 — with **exit 1 and real assertion errors**. Same job name, different signature.
- **Durations:** 18m18s · 21m36s · 22m02s · 28m17s · **1h0m0s**. main's shard 1 passes at ~18m09s.
- The 1h0m0s run hit `timeout-minutes: 55`; a timeout SIGTERMs the job and produces this exact signature. That run's shard-1 log is **absent from the attempts archive** — a hard-killed job never uploads logs.
- `ci.yml` sets `concurrency.cancel-in-progress: true` on `github.ref`, so any push SIGTERMs the in-flight run. Shard 1 is the longest job, so it is the most likely casualty. This plausibly explains the 18–28 min deaths.
- One run reached **624 of 812 files in 21m** — progressing, not hung, at that point.

### Prime suspect — UNVERIFIED

`vitest.memory-safe.ts` on this branch derives `MEMORY_SAFE_MAX_WORKERS` from cgroup + `MemAvailable` instead of `totalmem()`. If that reads low on a runner, workers drop to 1, the shard roughly doubles and approaches the 55-minute ceiling.

**Nobody has printed the resolved worker count on an actual runner.** That is the cheapest decisive measurement and it has not been taken.

### Suggested order

1. Echo `MEMORY_SAFE_MAX_WORKERS` in the shard step and read it off a runner — one CI run.
2. Pin `CLEO_TEST_MAX_WORKERS=2` in the unit-test job. CI parallelism should never depend on a heuristic that reads runner memory.
3. Re-run shard 1 with **no pushes in flight** to rule out `cancel-in-progress`; compare against a same-day main run.
4. If it still dies near 55 min, it is a hang — `--reporter=verbose` and find the last file entered but never completed.

### Also note

- `gate-10` (CLI barrel imports) fails identically on `main` from three files this PR never touches. Pre-existing — do **not** baseline it away.
- **Do not run the full suite on the developer machine.** It OOM-killed the owner's box three times in this session. Changed files only, `--maxWorkers=1`, or let CI do it.
- There is another active agent on **T12292** (epic T12256, unrelated). A tag publishes whatever is on `main` — confirm their work is landed or excluded before tagging.

